PR #2, opened 2017-01-19
Build basic OAuth authentication and download testcase information

everestmz:
cc @google/clusterfuzz 

The purpose of this CL is to build the initial mechanism aiming to download the relevant information for a specific testcase, before reproducing. As of right now, the flow is as follows:

`1) > clusterfuzz reproduce 12345`
`2) user is taken to the OAuth URL and given a verification token`
`3) user pastes token into the program when prompted`
`4) clusterfuzz API endpoint is hit to access test data, which is then printed onscreen`

Right now there is no local storage of any auth key, so the user must re-authenticate every time they wish to reproduce a test. The next CL will work on storing this token locally so users only need to re-authenticate manually when it expires.

Edit: Updating commits to have the correct author, to fix the CLA issue.
Edit2: Done


tanin47:
Please squash all commits into one commit.

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
```
tanin47: Sort alphabetically.

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
...
+import urllib
+import webbrowser
+import urlfetch
+
+
+def get_testcase_info(testcase_id, auth):
+  response = urlfetch.fetch(
+      url='https://staging-dot-cluster-fuzz.appspot.com/v2/testcase-detail/oauth?testcaseId=%s' % testcase_id,
```
tanin47: Move to a global constant and change `staging-dot-cluster-fuzz` to `cluster-fuzz`.

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
...
+  code = raw_input('Please enter your verification code:')
+
+  print 'Downloading testcase information...'
+
+  status, body, headers = get_testcase_info(
+      testcase_id,
+      'VerificationCode %s' % code)
+
+  if status != 200:
+    print 'Incorrect or expired verification code'
```
tanin47: Let's print the error message from the server here and mention that the verification code might be incorrect.

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
...
+    return
+
+  access_token = headers['x-clusterfuzz-authorization']
```
tanin47: Move x-clusterfuzz-authorization to a global constant.

requirements.txt
```diff
@@ -2,3 +2,4 @@ coveralls==1.1
 coverage==4.3.1
 mock==2.0.0
 twine==1.8.1
+urlfetch==1.0.2
```
tanin47: urlfetch is needed for the prod code (not development). I think it should be in setup.py. An example is here: https://github.com/coagulant/coveralls-python/blob/master/setup.py#L33

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,75 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
```
tanin47: Remove this line.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,75 @@
...
 import unittest
 
 from clusterfuzz.commands import reproduce
+import mock
```
tanin47: Move this import to above `import unittest`.

clusterfuzz/commands/reproduce.py
tanin47: Also, please remove this part. We already have the testcase info (when using auth=VerificationCode).

In the future, we'll save access_token locally to use in the future. (The snippet merely contains 2 examples for testing.)

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,75 @@
...
+
+    return self.response_code, response_body, response_headers
+
+  def setUp(self):
+    helpers.patch(self, [
+        'webbrowser.open',
+        'clusterfuzz.commands.reproduce.get_testcase_info',
+        '__builtin__.raw_input'])
+
+    self.mock.raw_input.side_effect = lambda _: '12345'
```
tanin47: Use `self.mock.raw_input.return_value = '12345'`

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,75 @@
...
+
+    self.mock.get_testcase_info.side_effect = self.clusterfuzz_response
```
tanin47: Use `self.mock.get_testcase_info.return_value` because it seems we don't care about the argument.

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
...
+
+  return response.status, json.loads(response.body), response.headers
+
+
+def generate_oauth_url():
+  client_id = '981641712411-sj50drhontt4m3gjc3hordjmpc7bn50f.apps.googleusercontent.com'
+  scope = 'email profile'
+  redirect_uri = 'urn:ietf:wg:oauth:2.0:oob'
+
+  url = 'https://accounts.google.com/o/oauth2/v2/auth?%s' % (
```
tanin47: This url is never changed. Let's make it a global constant and remove `generate_oauth_url()`.

clusterfuzz/commands/reproduce.py
tanin47: Let's raise an exception to exit instead of `return`.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,75 @@
...
+
+  def test_oauth_calls(self):
+    """Asserts that when verification is valid, API calls are made correctly."""
+
+    self.response_code = 200
+
+    reproduce.execute('1234', False)
+
+    self.mock.open.assert_has_calls([mock.call(
```
tanin47: `assert_has_calls` is insidious. `assert_has_calls([])` always succeeds. Therefore, here we should confirm that there's only one call.

In Python 2.7, we need to add `self.assertEqual(1, self.mock.open.call_count)`.

(This sounds like we need a helper like `assert_exact_calls(..)` to reduce code redundancy.)

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
...
+import webbrowser
+import urlfetch
+
+
+def get_testcase_info(testcase_id, auth):
+  response = urlfetch.fetch(
+      url='https://staging-dot-cluster-fuzz.appspot.com/v2/testcase-detail/oauth?testcaseId=%s' % testcase_id,
+      headers={'Authorization': auth})
+
+  return response.status, json.loads(response.body), response.headers
```
tanin47: This looks like a wrong abstraction of returned values.

`get_testcase_info` should return a testcase info, not status + body + headers.

We can check for the status 200 here and raise an exception to exit. And we can return a dict of testcase info instead.

clusterfuzz/commands/reproduce.py
```diff
@@ -12,7 +12,55 @@
...
+          'scope': scope,
+          'client_id': client_id,
+          'response_type': 'code',
+          'redirect_uri': redirect_uri}))
+  return url
+
 
 def execute(testcase_id, current):
   """Execute the reproduce command."""
+
```
tanin47: Remove this line.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,75 @@
...
 
+
 import unittest
 
 from clusterfuzz.commands import reproduce
+import mock
+from test import helpers
 
 
 class ExecuteTest(unittest.TestCase):
```
tanin47: After fixing the get_testcase_info's abstraction, there should be 2 tests: `ExecuteTest` and `GetTestcaseInfoTest`.

`ExecuteTest` should mock `get_testcase_info`; it doesn't test get_testcase_info.

`GetTestcaseInfoTest` should mock `urlfetch.fetch`.

everestmz:
Thanks! I'll get on the changes when I'm back at my desk. 

One question: some of your suggested changes (i.e. newlines, ordering of imports etc) were things I changed based on `glint`'s suggestions. Should I be using a different linter/style checker?

clusterfuzz/commands/reproduce.py
everestmz: Do you think there is a specific exception it makes sense to use? Or maybe write a clusterfuzz-tools specific one for this case?

tanin47:
> One question: some of your suggested changes (i.e. newlines, ordering of imports etc) were things I changed based on glint's suggestions. Should I be using a different linter/style checker?

Is this pylint? For now, let's follow glint.

clusterfuzz/commands/reproduce.py
tanin47: Good question. Ideally, I think we should have our own expected exception which is caught at the topmost level, so that we can hide the stacktrace and show only the error message. 

For now, let's make our own exception type first. Put the exception in `common.py` in the same directory where `main.py` is.

For the catching part, let's add that later in a separate CL, so this CL is small and focused.

For other unexpected exceptions, we want to print both the error message and the stacktrace.

everestmz:
>Is this pylint? For now, let's follow glint.

I used glint to style check the files.

mbarbella:
In this case gpylint is confused because it thinks urlfetch is part of the project rather than a library you depend on. It tries to block imports like:

import library_a
import library_b

import my_file_a
import my_file_b

everestmz:
Ah right, that makes sense. I'll change the files to reflect that. Thanks!

everestmz:
I think this is ready for another look - all suggestions & changes implemented.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,92 @@
...
+    self.assert_exact_calls(
+        self.mock.get_testcase_info,
+        [mock.call('1234', 'VerificationCode 12345')])
+
+class GetTestcaseInfoTest(ExtendedTestCase):
+  """Test get_testcase_info."""
+
+  def setUp(self):
+    helpers.patch(self, [
+        'json.loads',
```
tanin47: Let's not mock this.

And we can do `self.mock.fetch.return_value = mock.Mock(body=json.dumps({'id': 1234}, status=200, headers={something})`

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +12,92 @@
...
+        self.mock.loads])
+
+    self.assertEqual(response, return_body)
+
+  def test_incorrect_authorization(self):
+
+    self.mock.loads.return_value = {'status': 403, 'id': 'Bad_call'}
+    type(self.mock.fetch.return_value).status = mock.PropertyMock(return_value=403)
+
+    self.assertRaises(
```
tanin47: Change this one to:

```
with self.assertRaises(common.ClusterfuzzAuthError) as cm:
  reproduce.get_testcase_info('12345', 'Bad Auth')

self.assertIn('error-from-server', cm.exception.message)
```

Also, let's mock the error json properly (get one from the server).

everestmz:
Thanks, updated to only mock `urlfetch.fetch` in those tests.

clusterfuzz/common.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ClusterfuzzAuthError(Exception):
+  """An exception to deal with Clusterfuzz Authentication errors.
+
+  Makes the response dict available for inspection later on when
+  the exception is dealt with."""
+
+  def __init__(self, response):
+
+    message = "Error authenticating with Clusterfuzz\n%s" % str(response)
```
tanin47: Use single quotes (unless single quotes are present in the string). I wonder why pylint doesn't complain about this.

everestmz:
Rebasing, just making sure I caught all the lint errors first.

clusterfuzz/common.py
everestmz: Just saw this, will change that now

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -12,16 +13,103 @@
...
+
     reproduce.execute('1234', False)
-    self.assertTrue(True)
 
+    self.mock.open.assert_has_calls([mock.call(
+        reproduce.GOOGLE_OAUTH_URL,
+        autoraise=True,
+        new=1)])
+
+    self.assertEqual(1, self.mock.get_testcase_info.call_count)
```
tanin47: Whoops, remove this line.

clusterfuzz/commands/reproduce.py
```diff
@@ -1,3 +1,6 @@
+"""Module for the 'reproduce' command.
+
+Locally reproduces a testcase given a Cllusterfuzz ID."""
```
tanin47: Typo on ClusterFuzz

PR #3, opened 2017-01-20
[Not ready] Install pylint

tanin47:
We copy the pylintrc what gpylint uses. And we remove all google-specific stuffs.


tanin47:
Cancelled. I'll move to another PR.

PR #6, opened 2017-01-23
Clusterfuzz reproduce stores auth token locally

everestmz:
cc @google/clusterfuzz 

Clusterfuzz now stores an auth token in `~/.clusterfuzz` whenever test case info is pulled. This token is then used until it expires, so the user doesn't need to authenticate manually via the web browser every time they want to reproduce a crash. 

This PR also updates `circle.yml` to run `python setup.py install`. This was done to ensure all dependencies are installed before tests are run.


clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
         'response_type': 'code',
         'redirect_uri': 'urn:ietf:wg:oauth:2.0:oob'}))
 
+def auth_key_valid(auth_key):
+  """Checks whether auth_key is a valid authentication key."""
+
+  r = urlfetch.fetch(AUTH_CHECK_URL % auth_key)
+  return True if r.status == 200 else False
```
tanin47: return r.status == 200

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+
+def stored_auth_key():
```
tanin47: Let's name it `get_stored_auth_key()`.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+  """Checks whether there is a valid auth key stored locally."""
+  if not os.path.isfile(AUTH_KEY_FILE):
+    return None
+
+  permissions = int(oct(os.stat(AUTH_KEY_FILE).st_mode)[4:])
+  if permissions != 600 and permissions != 400:
+    return None
```
tanin47: Raise an exception saying that the file's permission is too permissive. This behaviour follows the SSH's behaviour.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+
+  with open(AUTH_KEY_FILE, 'r') as f:
+    auth_header = f.read()
+    if auth_header == '':
```
tanin47: Move this below block out of `with`. 

PS. auth_header is present outside `with`

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+      return None
+    key = auth_header[auth_header.find(' ')+1:]
```
tanin47: Move this to `auth_key_valid`'s body to separate responsibility clearly.

`get_stored_auth_key()` doesn't care what auth_header looks like. It returns the key when `auth_key_valid()` says that the key is valid.

clusterfuzz/commands/reproduce.py
```diff
@@ -45,6 +88,10 @@ def get_testcase_info(testcase_id, auth):
       url=CLUSTERFUZZ_TESTCASE_URL % testcase_id,
       headers={'Authorization': auth})
 
+  with open(AUTH_KEY_FILE, 'w+') as f:
```
tanin47: Let's remove `+`.

clusterfuzz/commands/reproduce.py
```diff
@@ -15,15 +15,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import json
 import urllib
 import webbrowser
+from stat import S_IREAD, S_IWUSR
```
tanin47: `import stat` and use `stat.S_IREAD` and `stat.S_IWUSR`. I'm surprised that pylint doesn't complain about this.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+    return None
+
+  with open(AUTH_KEY_FILE, 'r') as f:
+    auth_header = f.read()
+    if auth_header == '':
+      return None
+    key = auth_header[auth_header.find(' ')+1:]
+    return auth_header if auth_key_valid(key) else None
+
+def get_clusterfuzz_auth():
```
tanin47: Let's call this `get_auth_header()` and `get_stored_auth_header()` because we get and store the whole header.

clusterfuzz/commands/reproduce.py
```diff
@@ -57,14 +104,11 @@ def execute(testcase_id, current):
 
   print 'Reproduce %s (current=%s)' % (testcase_id, current)
 
-  webbrowser.open(GOOGLE_OAUTH_URL, new=1, autoraise=True)
-  code = raw_input('Please enter your verification code:')
+  code = get_clusterfuzz_auth()
```
tanin47: Rename `code` to `auth_header`

setup.py
```diff
@@ -30,4 +30,5 @@
     },
     install_requires=['urlfetch==1.0.2'],
     classifiers=[
-        'Programming Language :: Python :: 2.7'])
+        'Programming Language :: Python :: 2.7'],
+    data_files=[(os.path.expanduser('~/.clusterfuzz'), [])])
```
tanin47: Why do we need this? The pip package should already be able to write to any path it wants.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
         'response_type': 'code',
         'redirect_uri': 'urn:ietf:wg:oauth:2.0:oob'}))
 
+def auth_key_valid(auth_key):
```
tanin47: Rename it to `is_auth_header_valid`.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+  """Checks whether auth_key is a valid authentication key."""
+
+  r = urlfetch.fetch(AUTH_CHECK_URL % auth_key)
```
tanin47: We might do `return urlfetch.fetch(AUTH_CHECK_URL % auth_key).status == 200`.

I just want to confirm that it only returns 200 if the access token is valid right?

clusterfuzz/commands/reproduce.py
```diff
@@ -15,15 +15,21 @@
...
+import os
 import json
 import urllib
 import webbrowser
+from stat import S_IREAD, S_IWUSR
 import urlfetch
 
 from clusterfuzz.common import ClusterfuzzAuthError
 
+AUTH_KEY_FILE = os.path.expanduser('~/.clusterfuzz/auth_key')
```
tanin47: Rename this to `AUTH_HEADER`.

clusterfuzz/commands/reproduce.py
```diff
@@ -15,15 +15,21 @@
...
+AUTH_CHECK_URL = ('https://www.googleapis.com/oauth2/v3/tokeninfo'
```
tanin47: Rename this `TOKENINFO_URL`.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +38,43 @@
...
+    return auth_header if auth_key_valid(key) else None
+
+def get_clusterfuzz_auth():
+  """Returns a valid Clusterfuzz authorization key
+
+  If there is a valid authorization key stored locally
+  it will be used, otherwise the user will be prompted
+  for a new verification code."""
+
+  k = stored_auth_key()
```
tanin47: Rename k to `auth_header`.

clusterfuzz/commands/reproduce.py
tanin47: Let's change the word `key` to `header` because that's what we store.

clusterfuzz/commands/reproduce.py
tanin47: `key` here is actually `access_token`. So, let's name it `access_token`.

Also, I'd prefer a regex matching or something like `BEARER_PREFIX = 'Bearer '` and `auth_header[len(BEARER_PREFIX):]`

setup.py
everestmz: I thought that was necessary to create `~/.clusterfuzz` if it doesn't already exist so there are no problems making the file that holds the header.

tanin47: I see. I'd prefer creating the folder (if needed) when we need to store the header. 

Because currently the directory is only created for the user who installs the package; it can be that root installs the package, and some other users use it.

everestmz: Ok. I'll change the code so it checks for both directory & file presence.

clusterfuzz/commands/reproduce.py
everestmz: The + in `'w+'` is necessary so that if the file doesn't exist it's created, I believe.

everestmz:
All suggestions & changes addressed, thank you!

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +40,45 @@
...
+  it will be used, otherwise the user will be prompted
+  for a new verification code."""
+
+  header = get_stored_auth_header()
+  if header != None:
+    return header
+  else:
+    webbrowser.open(GOOGLE_OAUTH_URL, new=1, autoraise=True)
+    verification = raw_input('Please enter your verification code:')
+    code = 'VerificationCode %s' % verification
```
tanin47: Do return here. No need for `code`.

clusterfuzz/commands/reproduce.py
```diff
@@ -45,6 +92,13 @@ def get_testcase_info(testcase_id, auth):
       url=CLUSTERFUZZ_TESTCASE_URL % testcase_id,
       headers={'Authorization': auth})
 
+  if not os.path.exists(AUTH_HEADER_FILE.replace('/auth_header', '')):
```
tanin47: Use os.path.dirname to get the dir name.

I missed this part the first time. I think this logic should be in its own method (to be analogous to `get_stored_auth_header`).

Your initial thought to store the whole header (as opposed to access token) is a good one. Because clusterfuzz-tools should have the minimal knowledge about how auth header looks like.

clusterfuzz/commands/reproduce.py
```diff
@@ -15,15 +15,23 @@
...
 import urllib
 import webbrowser
+import stat
 import urlfetch
 
-from clusterfuzz.common import ClusterfuzzAuthError
+from clusterfuzz.common import (ClusterfuzzAuthError,
+                                ClusterfuzzPermissionsError)
 
+AUTH_HEADER_FILE = os.path.expanduser('~/.clusterfuzz/auth_header')
```
tanin47: I forgot. I think we should use `os.path.join('~', '.clusterfuzz', 'auth_header')` to support windows.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +40,45 @@
...
+  return urlfetch.fetch(TOKENINFO_URL % access_token).status == 200
+
+def get_stored_auth_header():
+  """Checks whether there is a valid auth key stored locally."""
+  if not os.path.isfile(AUTH_HEADER_FILE):
+    return None
+
+  permissions = int(oct(os.stat(AUTH_HEADER_FILE).st_mode)[4:])
+  if permissions != 600 and permissions != 400:
+    raise ClusterfuzzPermissionsError(
```
tanin47: This class is too general. We won't re-use this class for any other permission use case. We can remove `[400, 600]`. The first two arguments are good. And rename it too `PermissionTooPermissiveError`.

clusterfuzz/common.py
```diff
@@ -25,3 +25,21 @@ def __init__(self, response):
...
+  """An exception to deal with file permissions errors.
+
+  Stores the filename, the current permissions, and the
+  allowed permissions."""
+
+  def __init__(self, filename, current_perms, allowed_perms):
+    message_tuple = (filename,
+                     ', '.join([str(i) for i in allowed_perms]),
+                     str(current_perms))
+    message = ('Permissions error opening %s\nAllowed permissions: %s'
```
tanin47: We should say that the permission is too permissive. 

The message showing expected permission and current permission is super helpful.

clusterfuzz/commands/reproduce.py
```diff
@@ -45,6 +92,13 @@ def get_testcase_info(testcase_id, auth):
       url=CLUSTERFUZZ_TESTCASE_URL % testcase_id,
       headers={'Authorization': auth})
 
+  if not os.path.exists(AUTH_HEADER_FILE.replace('/auth_header', '')):
+    os.makedirs(AUTH_HEADER_FILE.replace('/auth_header', ''))
+
+  with open(AUTH_HEADER_FILE, 'w') as f:
+    f.write(response.headers[CLUSTERFUZZ_AUTH_HEADER])
+  os.chmod(AUTH_HEADER_FILE, stat.S_IWUSR|stat.S_IREAD)
```
tanin47: Why is it not S_IRUSR instead of s_IREAD?

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +40,45 @@
...
+    return False
+  access_token = auth_header[len(BEARER_PREFIX):]
+  return urlfetch.fetch(TOKENINFO_URL % access_token).status == 200
+
+def get_stored_auth_header():
+  """Checks whether there is a valid auth key stored locally."""
+  if not os.path.isfile(AUTH_HEADER_FILE):
+    return None
+
+  permissions = int(oct(os.stat(AUTH_HEADER_FILE).st_mode)[4:])
```
tanin47: This part offers very little semantics. Can we do something like:

```
can_group_access = bool(os.stat(AUTH_HEADER_FILE).st_mode & 0070))
can_other_access = bool(os.stat(AUTH_HEADER_FILE).st_mode & 0007))

if can_group_access or can_other_access:
  raise an exception
```

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -107,8 +104,138 @@ def test_incorrect_authorization(self):
...
+        autoraise=True,
+        new=1)])
+    self.assert_n_calls(1, [self.mock.raw_input])
+
+class GetStoredAuthHeaderTest(ExtendedTestCase):
+  """Tests the stored_auth_key method."""
+
+  def setUp(self):
+    helpers.patch(self, [
+        'os.path.isfile',
```
tanin47: Let's test this against the fake filesystem https://github.com/jmcgeheeiv/pyfakefs/, so that we don't need to mock st_mode directly.

Let's also test where we save auth_header and read it. Just to be sure the permissions are expected.

clusterfuzz/commands/reproduce.py
everestmz: Is the `expanduser` part still required, or does `.join` do that as well?

tanin47: I think expanduser is good. Let's keep it.

(`os.path.join` adds `/` between each string. On windows, it'll add `\` instead.)

clusterfuzz/commands/reproduce.py
everestmz: You're right, it should be `S_IRUSR`.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +41,55 @@
...
+  """Returns a valid Clusterfuzz authorization key
+
+  If there is a valid authorization key stored locally
+  it will be used, otherwise the user will be prompted
+  for a new verification code."""
+
+  header = get_stored_auth_header()
+  if header != None:
+    return header
+  else:
```
tanin47: +1

clusterfuzz/commands/reproduce.py
```diff
@@ -32,6 +40,45 @@
         'response_type': 'code',
         'redirect_uri': 'urn:ietf:wg:oauth:2.0:oob'}))
 
+def is_auth_header_valid(auth_header):
```
everestmz: I initially wrote this to store the entire header because the initial auth is done with the format `VerificationCode 12345`, so I was anticipating having to store that as well. Since that isn't the case now, I agree it should just store the token.

tanin47: Ideally, I'd like the client not to understand the header's content at all. The client simply receives the header, store, and later send the stored content to the server. So, there's no parsing needed in the client. 

This would enable the server to decide what the content should be. For example, when we support refresh_token, the server will send back access_token and refresh_token in the header (encoded in JSON maybe).

This begs the question whether the client should validate the access token. Maybe it can just catch exception and ask user to re-login if the header is invalid.

everestmz: It would make the tool more immune to future changes if all it did was store and retrieve the entire header. I'll edit the code to remove any specifics of the token format.

tanin47: As discussing offline, let's remove this function.

I'll modify the server to respond differently between a valid-access-token-but-no-access case and an invalid-access-token case. So, the client can react differently. This change can go in a different CL if you want.

tanin47: Ok, just checked the code. We are doing that.

If the access token is invalid, we return 401 (Unauthorized). If the access token is valid, but the user doesn't have access, then we return 403 (Access denied).

everestmz:
This has now been changed to reflect what we talked about yesterday, specifically that the client should not know anything about the structure of the authorization header. 

It needed a bit of a refactor because the old method `get_auth_header` could no longer guarantee a working token, but I think this is more in the spirit of what we were trying to achieve.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,39 +40,72 @@
...
+    return None
 
-def get_testcase_info(testcase_id, auth):
+  can_group_access = bool(os.stat(AUTH_HEADER_FILE).st_mode & 0070)
+  can_other_access = bool(os.stat(AUTH_HEADER_FILE).st_mode & 0007)
+
+  if can_group_access or can_other_access:
+    raise PermissionsTooPermissiveError(
+        AUTH_HEADER_FILE,
+        int(oct(os.stat(AUTH_HEADER_FILE).st_mode)[4:]))
```
tanin47: `oct(os.stat(AUTH_HEADER_FILE).st_mode & 0777)`

Changing it to int becomes unrecognisable number.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,39 +40,72 @@
...
+  """Stores 'auth_header' locally for future access."""
+
+  if not os.path.exists(os.path.dirname(AUTH_HEADER_FILE)):
+    os.makedirs(os.path.dirname(AUTH_HEADER_FILE))
+
+  with open(AUTH_HEADER_FILE, 'w') as f:
+    f.write(auth_header)
+  os.chmod(AUTH_HEADER_FILE, stat.S_IWUSR|stat.S_IRUSR)
+
+def get_testcase_url(testcase_id, auth):
```
tanin47: The method's name is strange. I think it would be ok not to make this a method, now that we call any method in a single place.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,39 +40,72 @@
...
+  can_other_access = bool(os.stat(AUTH_HEADER_FILE).st_mode & 0007)
+
+  if can_group_access or can_other_access:
+    raise PermissionsTooPermissiveError(
+        AUTH_HEADER_FILE,
+        int(oct(os.stat(AUTH_HEADER_FILE).st_mode)[4:]))
+
+  with open(AUTH_HEADER_FILE, 'r') as f:
+    return f.read()
+
```
tanin47: We use 2 lines to separate top-level methods.

clusterfuzz/commands/reproduce.py
```diff
@@ -32,39 +40,72 @@
...
   Returns a dictionary with the JSON response if the
   authentication is successful, or throws a
   ClusterfuzzAuthError otherwise.
   """
-
-  response = urlfetch.fetch(
-      url=CLUSTERFUZZ_TESTCASE_URL % testcase_id,
-      headers={'Authorization': auth})
+  header = get_stored_auth_header()
+  response = get_testcase_url(testcase_id, header)
```
tanin47: This part looks a bit strange. Let's combine them into a for-loop, so that we call any method in a single place.

```
response = None
for _ in range(2):
  if not header or (response and response.status == 401):
    header = get_verification_header()
  response = get_testcase_url(...)
```

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -42,58 +61,80 @@ class ExecuteTest(ExtendedTestCase):
...
+    self.mock.get_stored_auth_header.return_value = None
+    self.mock.get_verification_header.return_value = 'VerificationCode 12345'
+    self.mock.get_testcase_url.return_value = mock.Mock(
+        status=200,
+        body=json.dumps(response_dict),
+        headers=response_headers)
+
+    response = reproduce.get_testcase_info('12345')
+
+    self.assert_n_calls(1, [
```
tanin47: Let's verify the arguments that were called as well, e.g. the correct argument of store_auth_header is important.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -42,58 +61,80 @@ class ExecuteTest(ExtendedTestCase):
...
         status=200,
-        body=json.dumps(response_dict))
+        body=json.dumps(response_dict),
+        headers=response_headers)
 
-    response = reproduce.get_testcase_info('12345', 'Test Auth')
+    response = reproduce.get_testcase_info('12345')
 
-    self.assert_n_calls(1, [self.mock.fetch])
+    self.assert_n_calls(1, [
```
tanin47: Verify the called arguments.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -42,58 +61,80 @@ class ExecuteTest(ExtendedTestCase):
...
+        self.mock.get_testcase_url,
+        self.mock.get_stored_auth_header,
+        self.mock.store_auth_header])
+    self.assertEqual(response, response_dict)
+
+  def test_correct_verification_auth(self):
```
tanin47: Add one test for bad stored header but logging in is successful.

```
self.mock.get_testcase_url.side_effect = [
  mock.Mock(status=401),
  mock.Mock(status=200, ...)
]
```

Use `side_effect` to make the mock return 2 different values when called twice.

everestmz: Thanks, I didn't know about that one

clusterfuzz/commands/reproduce.py
```diff
@@ -32,39 +40,75 @@
...
   """
 
-  response = urlfetch.fetch(
+  header = get_stored_auth_header()
+  response = None
+  for _ in range(2):
+    if not header or (response and response.status == 401):
+      header = get_verification_header()
+      response = None
+    response = response if response else urlfetch.fetch(
```
everestmz: I had to edit the loop a little, otherwise it would try to use the same verification code twice, causing an error. Now it only accesses the API if `response == None`.

tanin47: Oh, you're right. It's better to `break` when response is 200. And we can remove `response = None`.

test/clusterfuzz/commands/reproduce_test.py
tanin47: `get_stored_auth_header` and `get_verification_header`'s arguments should be verified with e.g. `self.assert_exact_calls(self.mock.store_auth_header, [mock.call()])`.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -42,58 +61,119 @@ class ExecuteTest(ExtendedTestCase):
...
+        mock.Mock(status=200,
+                  body=json.dumps(response_dict),
+                  headers=response_headers)]
+    self.mock.get_stored_auth_header.return_value = 'Bearer 12345'
+    self.mock.get_verification_header.return_value = 'VerificationCode 12345'
+
+    response = reproduce.get_testcase_info('12345')
+
+    self.assert_n_calls(1, [
+        self.mock.get_stored_auth_header,
```
tanin47: Verify the arguments of these methods too.

everestmz: `get_stored_auth_header` and `get_verification_header` don't take any arguments

tanin47: Yes, that's what we verify e.g. `mock.call()`.

everestmz: Oh I understand, got it.

everestmz:
All tests updated to verify method calls correctly

clusterfuzz/commands/reproduce.py
```diff
@@ -32,39 +40,76 @@
...
   ClusterfuzzAuthError otherwise.
   """
 
-  response = urlfetch.fetch(
-      url=CLUSTERFUZZ_TESTCASE_URL % testcase_id,
-      headers={'Authorization': auth})
+  header = get_stored_auth_header()
+  response = None
+  for _ in range(2):
+    if response and response.status == 200:
```
tanin47: Move this to below urlfetch and we can drop the first condition `response` and only have `response.status == 200`

tanin47:
Awesome. I just tested. It works well!

PR #7, opened 2017-01-25
Switch build shield to circleci

tanin47:


everestmz:
Was just about to push a branch to do this myself :smile:

PR #8, opened 2017-01-27
Reproduce gets the chrome revision, converts to SHA, and checks out the branch

everestmz:
cc: @google/clusterfuzz 

Reproduce uses the `CHROME_SRC` environment variable to decide where the chrome source code is. 
It then finds the Chrome release number from within the downloaded testcase data, and uses the API to convert this revision number into a commit sha.
Finally, it runs a `git fetch && git checkout <sha>` to checkout the correct build version.


clusterfuzz/commands/reproduce.py
```diff
@@ -97,12 +97,59 @@ def get_testcase_info(testcase_id):
...
+
+
+def sha_from_revision(revision_number):
+  """Converts a chrome revision number to it corresponding git sha."""
+
+  response = urlfetch.fetch(build_revision_to_sha_url(revision_number))
+  return json.loads(response.body)['git_sha']
+
+
+def ask_question(question, default='y'):
```
everestmz: This method may not be needed, but it depends on how many times we ask questions in the future so I think it makes sense to abstract this functionality into a method for now.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -29,6 +29,11 @@ class ExtendedTestCase(fake_filesystem_unittest.TestCase):
   """An extended version of TestCase with extra methods for fine-grained method
   call assertions."""
 
+  def setup_fake_environment(self):
```
everestmz: I was trying to think of a better way to do this, and will continue to. It's a little difficult as `clusterfuzz.commands.reproduce.CHROME_SOURCE` is defined before any of the `setUp` methods are run, meaning mocking `os.environ` doesn't work too well. 

I considered adding ENV variables to the `circle.yml` but then local tests aren't guaranteed to run. Another option may be to reload the module after mocking `os.environ`, so it reloads `CHROME_SOURCE`, but I would like to have another opinion on the best way to fake environment variables.

clusterfuzz/commands/reproduce.py
```diff
@@ -97,12 +97,59 @@ def get_testcase_info(testcase_id):
...
+  is returned as either True or False."""
+
+  default_list = ['y', 'n', ''] if default else ['y', 'n']
+  defaults = '[y/n] '
+  if default == 'y':
+    defaults = '[Y/n] '
+  elif default == 'n':
+    defaults = '[y/N] '
+
+  answer = raw_input(''.join([question, ' ', defaults])).lower()
```
tanin47: Use string interpolation and add colon. `'%s %s:' % (question, defaults)`

We might wanna add `.strip()` as well.

clusterfuzz/common.py
```diff
@@ -43,3 +48,40 @@ def __init__(self, filename, current_permissions):
...
     self.filename = filename
     self.current_permissions = current_permissions
+
+
+def execute(command,
+            cwd,
+            print_output=True,
+            exit_on_error=True,
+            extra_environments=None):
+  """Execute a bash command."""
```
tanin47: We need to write tests on this method. Let's trim it down to what we really need. So, we don't have to write too many tests.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -282,3 +297,132 @@ def test_returns_correct_header(self):
...
+  def test_yes_default(self):
+    """Tests functionality with yes as default."""
+
+    self.mock.raw_input.side_effect = ['y', 'n', '']
+
+    r1 = reproduce.ask_question('A question')
+    r2 = reproduce.ask_question('A question')
+    r3 = reproduce.ask_question('A question')
+
+    self.assertEqual(r1, r3, True)
```
tanin47: I don't think we can do this. This compares r1 to r3; They happen to be equal. The third argument is the error message.

Use `self.assertTrue(reproduce.ask('A question'))`

clusterfuzz/commands/reproduce.py
tanin47: And I think we want to move this method to common.py

clusterfuzz/commands/reproduce.py
```diff
@@ -97,12 +97,59 @@ def get_testcase_info(testcase_id):
...
+
+
+def ask_question(question, default='y'):
+  """Asks the user a question and returns their answer.
+
+  default can either be 'y', 'n', or None. Answer
+  is returned as either True or False."""
+
+  default_list = ['y', 'n', ''] if default else ['y', 'n']
+  defaults = '[y/n] '
```
tanin47: Remove the trailing space.

And do `'[y/n'].replace(default, default.upper())`

clusterfuzz/commands/reproduce.py
```diff
@@ -97,12 +97,59 @@ def get_testcase_info(testcase_id):
...
+  return json.loads(response.body)['git_sha']
+
+
+def ask_question(question, default='y'):
+  """Asks the user a question and returns their answer.
+
+  default can either be 'y', 'n', or None. Answer
+  is returned as either True or False."""
+
+  default_list = ['y', 'n', ''] if default else ['y', 'n']
```
tanin47: This is not a default list. I feel it's more like an acceptable answer set.

And to avoid repeating `'y', 'n'`, let's do 

```
accepts =  ['y', 'n']
if default:
  accepts += ['']
```

clusterfuzz/common.py
everestmz: I was thinking about that - since we borrowed it from the clusterfuzz repo, are there tests we could use from there to make sure no functionality is missed? I agree stripping it down is a good idea, there may be things we don't need.

tanin47:
Sadly, we don't have tests for this method. Let's strip it down.

clusterfuzz/common.py
tanin47: Sadly, we don't have tests for this method. Let's strip it down.

everestmz:
Updated with tests for `common.execute` to keep test coverage at :100:

clusterfuzz/commands/reproduce.py
```diff
@@ -97,19 +96,51 @@ def get_testcase_info(testcase_id):
...
+  return json.loads(response.body)['git_sha']
+
+
+def checkout_chrome_by_sha(sha, chrome_source):
+  """Checks out the correct Chrome revision."""
+
+  command = 'git fetch && git checkout %s' % sha
+  if not common.confirm('Proceed with the following command:\n%s in %s?' %
+                        (command, chrome_source)):
+    exit()
```
tanin47: I didn't know about `exit()` before. This person seems to suggest that we should use `sys.exit()` instead. http://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python

`exit()` is meant for interactive shell.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -282,3 +262,80 @@ def test_returns_correct_header(self):
...
+  def setUp(self):
+    self.setup_fake_environment()
+    helpers.patch(self, [
+        'clusterfuzz.common.execute',
+        'clusterfuzz.common.confirm'])
+    self.chrome_source = '/usr/local/google/home/user/repos/chromium/src'
+    self.command = ('git fetch && git checkout 1a2s3d4f'
+                    ' in %s' % self.chrome_source)
+
+  def test_answer_no(self):
```
tanin47: If we have a separate method for `check_confirm` where it either does nothing or `exit()`.

We would remove the if condition from `checkout_chrome_by_sha`, and we wouldn't need `test_answer_no`. 

But we need two tests (yes and no) for `check_confirm`. But `check_confirm` can be re-used and is easier to test because it has fewer dependencies.

test/clusterfuzz/common_test.py
```diff
@@ -0,0 +1,138 @@
...
+  def test_process_run_fails(self):
+    """Test execute when the process does not run successfully."""
+
+    return_code = 1
+
+    self.run_popen_assertions(return_code, True, True)
+    self.run_popen_assertions(return_code, True, False)
+    self.run_popen_assertions(return_code, False, True)
+    self.run_popen_assertions(return_code, False, False)
+    self.mock.Popen.assert_has_calls([mock.call(
```
tanin47: Do the same thing here.

test/clusterfuzz/common_test.py
```diff
@@ -0,0 +1,138 @@
...
+    self.assertEqual(return_code, None if will_exit else code)
+    self.assertEqual(returned_lines, None if will_exit else self.lines)
+    self.assert_exact_calls(self.mock.Popen.return_value.wait, [mock.call()])
+
+  def test_process_runs_successfully(self):
+    """Test execute when the process successfully runs."""
+
+    return_code = 0
+
+    self.run_popen_assertions(return_code, True, True)
```
tanin47: This is clever! We can make it shorter and offer more semantics:

```
for print_out in [True, False]:
  for exit_on_error in [True, False]:
     self.run_popen_assertions(return_code, True, False)
```

In `run_popen_assertions`, you can do:

```
self.mock.Popen.reset_mock()  # this reset the calls.
...
self.mock.assert_exact_calls(self.mock.Popen, [mock.call(..)])
```

clusterfuzz/commands/reproduce.py
```diff
@@ -97,19 +96,51 @@ def get_testcase_info(testcase_id):
...
+
+
 def execute(testcase_id, current):
   """Execute the reproduce command."""
 
   print 'Reproduce %s (current=%s)' % (testcase_id, current)
   print 'Downloading testcase information...'
 
   response = get_testcase_info(testcase_id)
+  chrome_source = os.environ.get('CHROME_SRC')
```
tanin47: Let's do `os.environ['CHROME_SRC']`, so that it fails when the env is not present.

test/helpers.py
```diff
@@ -12,14 +13,56 @@
...
+    self.setUpPyfakefs()
+    self.clusterfuzz_dir = os.path.expanduser(os.path.join(
+        '~', '.clusterfuzz'))
+    self.auth_header_file = os.path.join(self.clusterfuzz_dir,
+                                         'auth_header')
+
+  def assert_file_permissions(self, filename, permissions):
+    """Assert that 'filename' has specific permissions"""
+
+    self.assertEqual(int(oct(os.stat(filename).st_mode)[4:]),
```
tanin47: use 0777 as we use in the code previously.

everestmz: I was using this method as it outputs the status code as the octal (i.e 755), which allows the user to run `assert_file_permissions(filename, 755)` rather than having to go through any conversions. Or is there a better way to achieve this?

tanin47: That's sensible. But I'd like to change `os.stat(filename).st_mode)[4:]` to `os.stat(AUTH_HEADER_FILE).st_mode & 0777` instead.

everestmz:
Thank you for the feedback, all updated.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -282,3 +263,82 @@ def test_returns_correct_header(self):
...
         autoraise=True)])
     self.assertEqual(response, 'VerificationCode 12345')
+
+
+class BuildRevisionToShaUrlTest(helpers.ExtendedTestCase):
+  """Tests the build_revision_to_sha_url method."""
+
+  def setUp(self):
+
+    helpers.patch(self, [
```
tanin47: New line is not needed. We generally want to minimize the vertical space.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -19,61 +19,42 @@
...
         'crash_type': 'Bad Crash',
-        'crash_state': ['halted']}
+        'crash_state': ['halted'],
+        'crash_revision': '123456'}
 
   def test_grab_data(self):
-    """Asserts that the testcase data is grabbed correctly."""
+    """Ensures all method calls are made correctly."""
+    self.mock.sha_from_revision.return_value = '1a2s3d4f'
+    with mock.patch.dict(
```
tanin47: Can this be moved to `setUp`. I'm not sure whether it will work.

```
self.environ = {}
patcher = mock.patch.dict('os.environ', self.environ)
patcher.start()
self.addCleanup(patcher.stop)
```

Then, in a test, we can mock:

```
self.environ['CHROME_SRC'] = 'something'
```

Good use of `mock.patch.dict`. We never used it before.

everestmz: It looks as though `patch.dict` patches when it is run so any ENV vars need to be in self.environ before the patch method is run. It can still be done in setup though.

PR #9, opened 2017-02-01
Clusterfuzz successfully builds the Chrome revision using GOMA

everestmz:
cc @google/clusterfuzz 

Pushing this without tests written first in case there are some larger design changes to make. This CL does a fair bit. 

In `common.py`, updates `execute` to include the capability to print shell output live (so the user isn't waiting until the end of a build before they see any output), as well as the ability to add input to stdin (required to build GOMA).

EDIT: after discussion, this script no longer auto-builds GOMA and therefore `common.execute` no longer needs stdin capability.

For the `reproduce` command, adds the functionality necessary to download, configure and build chrome. Also includes methods to install and set up GOMA prior to running. In addition, added a couple of checks for situations where commands have already been run (e.g. if the correct commit is already checked out, don't check it out again). This means that re-running the command for the same crash goes through the entire process without requiring any input from the user.


clusterfuzz/commands/reproduce.py
```diff
@@ -130,12 +140,125 @@ def check_confirm(question):
...
+
+  os.makedirs(GOMA_DIR)
+  os.environ['GOMA_DIR'] = GOMA_DIR
+  common.execute(('curl https://clients5.google.com/cxx-compiler-service/'
+                  'download/goma_ctl.py -o goma_ctl.py'),
+                 GOMA_DIR)
+  #TODO: Make this GOMA installer call platform agnostic
+  common.execute('python goma_ctl.py update',
+                 GOMA_DIR,
+                 input_string='1', #This needs to be changed based on OS
```
tanin47: Looking at `goma_ctl.py`'s code. There's a way to inject platform directly. On line 1174, we can set the env `PLATFORM` to be `goobuntu`.

I'm a bit concerned that installing goma is on a wrong abstraction. We don't usually do one-time installation (e.g. cloning chrome, `pip install -r requirements.txt`). We can error out and ask users to run these commands to install it themselves. But checking for goma installation is good.

clusterfuzz/common.py
```diff
@@ -64,10 +66,16 @@ def _print(s):
...
       shell=True,
-      stdout=subprocess.PIPE,
-      stderr=subprocess.STDOUT,
+      stdin=subprocess.PIPE,
+      stdout=None if live_output else subprocess.PIPE,
+      stderr=None if live_output else subprocess.STDOUT,
       cwd=cwd)
-  for line in iter(proc.stdout.readline, b''):
+
+  if live_output:
```
tanin47: Hmmm... this is strange. The live output should already work. Could you test it with `execute('yes')` to see if we see the output as it prints.

clusterfuzz/commands/reproduce.py
```diff
@@ -130,12 +140,125 @@ def check_confirm(question):
...
+  if os.path.isfile(args_gn_location):
+    os.remove(args_gn_location)
+
+  common.execute('gn gen out/clusterfuzz_%s' % testcase_id, chrome_source)
+  shutil.copy(
+      os.path.join(get_build_directory(testcase_id), 'args.gn'),
+      args_gn_location)
+
+  goma_location = install_goma()
+  startup_goma(goma_location)
```
tanin47: We can combine these two methods and call `ensure_goma(goma_location)`.

clusterfuzz/commands/reproduce.py
```diff
@@ -130,12 +140,125 @@ def check_confirm(question):
...
+  if os.path.exists(build_dir):
+    return build_dir
+
+  print 'Downloading Chrome build data...'
+
+  if not os.path.exists(CLUSTERFUZZ_BUILDS_DIR):
+    os.makedirs(CLUSTERFUZZ_BUILDS_DIR)
+
+  common.execute(
+      'wget %s' % build_url.replace('.cloud.google.', '.googleapis.'),
```
everestmz: Even when changing the filename to a temp name, the zip itself still unzips into `asan-linux-debug.....` format. As far as I can tell looking at `ZipFile` there isn't a way to rename the directory before it is unzipped, so we may have to live with using the given name when extracting.

tanin47: That's true. We do need to know the folder name.

everestmz:
Suggestions fixed, and the tool can successfully build d8 which can be used to (manually) successfully reproduce a crash. I will start writing the tests now.

clusterfuzz/common.py
```diff
@@ -52,24 +52,36 @@ def __init__(self, filename, current_permissions):
...
+    #return
+
+  #stdoutdata, _ = proc.communicate(input=input_string)
+  #for line in stdoutdata.split('\n'):
+    #_print('| %s' % line.rstrip())
+    #lines.append(line)
+  for byte in iter(lambda: proc.stdout.read(1), b''):
+    sys.stdout.write(byte)
+    output += byte
+  lines = output.split('\n')
```
tanin47: no need to split it. We don't use line for anything. We return the whole string anyway.

clusterfuzz/common.py
```diff
@@ -52,24 +52,36 @@ def __init__(self, filename, current_permissions):
...
-    lines.append(line)
+
+  #if live_output:
+    #return
+
+  #stdoutdata, _ = proc.communicate(input=input_string)
+  #for line in stdoutdata.split('\n'):
+    #_print('| %s' % line.rstrip())
+    #lines.append(line)
+  for byte in iter(lambda: proc.stdout.read(1), b''):
```
tanin47: That is interesting. Thank you for figuring this out. I'm gonna copy it to clusterfuzz.

clusterfuzz/commands/reproduce.py
```diff
@@ -76,31 +81,36 @@ def store_auth_header(auth_header):
...
   if response.status != 200:
-    raise common.ClusterfuzzAuthError(body)
-
+    raise common.ClusterfuzzAuthError(response.body)
   store_auth_header(response.headers[CLUSTERFUZZ_AUTH_HEADER])
-  return body
+
+  return response
+
+def get_testcase_info(testcase_id):
```
tanin47: I feel that this method doesn't do much.

However, it explains what the logic does (as opposed to add a comment explaining it). So, keeping it is good.

clusterfuzz/common.py
everestmz: I'm not sure if you've moved this code to clusterfuzz yet, but if you haven't then wait a little bit - as of now it doesn't put the `| ` in front of every line that it did before. I will update this and push it with the tests.

tanin47: I can live with that. `I` is a poorly-thought design anyway.

On a larger scheme, we want user to understand what we run and understand the large amount of stdout. I'm not sure what the best UX/UI to achieve that. You can ignore that part for now.

everestmz: Ok, sounds good. I'll leave that out for now then.

everestmz:
Updated with tests & cleaned up all comments/extra pieces of code - local d8 builds working.

clusterfuzz/commands/reproduce.py
```diff
@@ -130,12 +139,114 @@ def check_confirm(question):
...
+  os.remove(saved_file)
+  os.rename(os.path.join(CLUSTERFUZZ_BUILDS_DIR, os.path.splitext(filename)[0]),
+            build_dir)
+
+
+def ensure_goma():
+  """Ensures GOMA is installed and ready for use, and starts it."""
+
+  goma_dir = os.environ.get('GOMA_DIR', GOMA_DIR)
+  if not (os.path.exists(goma_dir) and os.path.isfile(os.path.join(
```
tanin47: I wonder if we could just do `os.path.isfile` here.

clusterfuzz/commands/reproduce.py
```diff
@@ -130,12 +139,114 @@ def check_confirm(question):
...
+  testcase_source_dir = get_out_dir(chrome_source, testcase_id)
+  print 'Building Chrome revision %i in %s' % (
+      revision_number,
+      testcase_source_dir)
+
+  goma_dir = ensure_goma()
+  setup_gn_args(testcase_source_dir, testcase_id, chrome_source, goma_dir)
+
+  goma_cores = 10 * multiprocessing.cpu_count()
+  common.execute('GYP_DEFINES=asan=1 gclient runhooks', chrome_source)
```
tanin47: @mbarbella mentioned to me that we should have been able to build v8 inside chrome's source code directly. Actually, it should have worked with any other component.

But let's investigate that after we finish the symbolization step.

clusterfuzz/common.py
```diff
@@ -59,25 +68,27 @@ def _print(s):
...
+      stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.STDOUT,
       cwd=cwd)
-  for line in iter(proc.stdout.readline, b''):
-    _print('| %s' % line.rstrip())
-    lines.append(line)
+
+  for byte in iter(lambda: proc.stdout.read(1), b''):
+    sys.stdout.write(byte)
```
tanin47: This line doesn't respect `print_output`.

I think we should re-use `_print`. Maybe rename it to `_write` and and use `sys.stdout.write(byte)` inside instead. Hmm but maybe `sys.stdout.write(byte)` doesn't work with string.

Or we can just guard this line with `if print_output`.

clusterfuzz/commands/reproduce.py
```diff
@@ -130,12 +139,114 @@ def check_confirm(question):
 def checkout_chrome_by_sha(sha, chrome_source):
   """Checks out the correct Chrome revision."""
 
+  current_sha = common.execute('git rev-parse HEAD',
```
tanin47: Use `_, current_sha = common.execute(..)` instead of `[1]`.

clusterfuzz/common.py
```diff
@@ -59,25 +68,27 @@ def _print(s):
       print s
 
   _print('Running: %s' % command)
-  lines = []
+  output = ''
 
   proc = subprocess.Popen(
       command,
       shell=True,
+      stdin=subprocess.PIPE,
```
tanin47: Remove. We don't use it anymore.

test/clusterfuzz/common_test.py
```diff
@@ -106,6 +106,7 @@ def run_popen_assertions(self, code, print_out=True, exit_on_err=True):
     self.assert_exact_calls(self.mock.Popen, [mock.call(
         'cmd',
         shell=True,
+        stdin=subprocess.PIPE,
```
tanin47: Remove here as well.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -331,14 +337,182 @@ def setUp(self):
...
+  def test_get_build_data(self):
+    """Tests extracting, moving and renaming the build data.."""
+
+    os.makedirs(self.clusterfuzz_dir)
+    cf_builds_dir = os.path.join(self.clusterfuzz_dir, 'builds')
+
+    with open(os.path.join(self.clusterfuzz_dir, 'args.gn'), 'w') as f:
+      f.write('use_goma = True')
+    fakezip = zipfile.ZipFile(
+        os.path.join(self.clusterfuzz_dir, 'abc.zip'),
```
tanin47: Try to save vertical space. No need for new line here.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -331,14 +337,182 @@ def setUp(self):
...
+
+    self.assert_exact_calls(self.mock.execute, [mock.call(
+        'gsutil cp gs://abc.zip .',
+        self.clusterfuzz_dir)])
+    self.assertFalse(
+        os.path.isfile(os.path.join(self.clusterfuzz_dir, 'abc.zip')))
+    self.assertTrue(os.path.isdir(
+        os.path.join(cf_builds_dir, '12345_build')))
+    self.assertTrue(os.path.isfile(os.path.join(
+        cf_builds_dir,
```
tanin47: Save vertical space. Reduce the next 3 lines to one line.

clusterfuzz/common.py
```diff
@@ -59,25 +68,27 @@ def _print(s):
...
       stdout=subprocess.PIPE,
       stderr=subprocess.STDOUT,
       cwd=cwd)
-  for line in iter(proc.stdout.readline, b''):
-    _print('| %s' % line.rstrip())
-    lines.append(line)
+
+  for byte in iter(lambda: proc.stdout.read(1), b''):
+    sys.stdout.write(byte)
+    output += byte
```
tanin47: I'm surprised that we can do this. It might fail when there's a unicode character. Since it is working, let's keep it.

clusterfuzz/common.py
everestmz: I could add a handler in `_print` that either `print`s or `sys.stdout.write`s based on the type of the input to `_print`.

clusterfuzz/commands/reproduce.py
everestmz: It's definitely weird that we can't build from the exact same repo as soon as it is a sub-repo of chromium. I'm going to try and re-clone chromium again sometime, and see if working with a fresh repo produces any different results.

clusterfuzz/commands/reproduce.py
everestmz: You're right, that would be a nicer line. The first part is redundant.

clusterfuzz/common.py
tanin47: Hmm.. I'm not sure which way makes it the most beautiful.

print and sys.stdout.write seem to do two different things and take different inputs.

test/clusterfuzz/commands/reproduce_test.py
everestmz: All on one line it goes 1 char over 80 long, and the pylint complains

clusterfuzz/common.py
everestmz: I'll just guard the line with an if statement, it will keep things clear.

test/clusterfuzz/commands/reproduce_test.py
tanin47:
```
fakezip = zipfile.ZipFile(
    os.path.join(self.clusterfuzz_dir, 'abc.zip'), 'w')
```

I mean like the above.

clusterfuzz/common.py
tanin47: Sound good. And, if we have a more beautiful way of doing this, we can do it later.

everestmz:
Thank you - all suggestions addressed!

tanin47:
I tested in on my machine. It works as expected. It asked for Goma. After installing Goma, it worked on first try. Awesome job!

PR #10, opened 2017-02-08
Clusterfuzz-tools can successfully reproduce some d8 testcases

everestmz:
cc: @google/clusterfuzz 

Reproduce now successfully reproduces some d8 testcases. A list of testcases that have been tried with clusterfuzz-tools will be kept up to date in [this document.](https://docs.google.com/a/google.com/document/d/17QdRwHHNYoEVhQua3t4fHLMCh_BMgCG2VUAdzs3ameg/edit?usp=sharing)

This CL allows for the downloading and reproduction of certain d8 testcases. As of now, d8 specifics are hardcoded, but this will change as I add support for more build targets & testcases.


clusterfuzz/commands/reproduce.py
```diff
@@ -245,17 +256,71 @@ def build_chrome(revision_number, testcase_id, chrome_source):
...
+      auth_header, CLUSTERFUZZ_TESTCASE_URL % testcase_id)
+  common.execute(command, testcase_dir)
+
+  return filename
+
+
+def set_up_environment(stacktrace_lines):
+  """Sets up the environment by parsing stacktrace lines."""
+
+  new_env = os.environ.copy()
```
tanin47: Why do we need to copy our current env? I think it would be better to use a fresh env.

clusterfuzz/commands/reproduce.py
```diff
@@ -245,17 +256,71 @@ def build_chrome(revision_number, testcase_id, chrome_source):
...
+  stacktrace_lines = [l['content']  for l in stacktrace_lines]
+  for l in stacktrace_lines:
+    if '[Environment] ' not in l:
+      continue
+    l = l.replace('[Environment] ', '')
+    name, value = l.split(' = ')
```
tanin47: Let's split by `=` and do `.strip()` on both `name` and `value`.

And thanks for writing part. I know we should have provided this info in the JSON from clusterfuzz instead. But that requires refactoring and time.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -513,3 +522,103 @@ def test_correct_calls(self):
...
+  def test_get_env_from_stacktrace(self):
+    """Tests grabbing the os environment and adding the new vars from lines."""
+    stacktrace_lines = [
+        {'content': '[Environment] ASAN_OPTIONS = option1=true:option2=false'},
+        {'content': '======== Stacktrace ======'},
+        {'content':
+         '[Environment] GOMA_OPTIONS = goma_dir=/goma/dir:use_goma=true'}]
+    correct_env = os.environ.copy()
+    correct_env['ASAN_OPTIONS'] = 'option1=true:option2=false'
+    correct_env['GOMA_OPTIONS'] = 'goma_dir=/goma/dir:use_goma=true'
```
tanin47: Good job spotting this. I didn't know we could have multiple lines of `[Environment]`.

clusterfuzz/commands/reproduce.py
```diff
@@ -265,6 +330,8 @@ def execute(testcase_id, current):
       testcase_id)
   build_chrome(crash_revision, testcase_id, chrome_source)
 
-  print 'Testcase ID: %i' % testcase_id
-  print 'Crash Type: %s' % response['crash_type']
-  print 'Crash State: %s' % ', '.join(response['crash_state'])
+  reproduction_args = get_reproduction_args(response)
+  testcase_file = download_testcase_file(testcase_id)
+  env = set_up_environment(response['crash_stacktrace']['lines'])
+  reproduce_crash(testcase_id, testcase_file,
```
tanin47: I feel we should pass the whole testcase into reproduce_crash. But since the testcase is a json, we might want to convert json into a class out of it, so that we can encapsulate behaviours. When converting, we only take what we need.

For example, `get_reproduction_args` should be encapsulated inside the testcase class. Maybe `get_testcase_dir` can be in the testcase class as well.

Let's refactor this in the next CL... because this CL looks good.

clusterfuzz/commands/reproduce.py
everestmz: I did it because, as per `Popen`'s documentation, this env is taken "instead of inheriting the current process’ environment, which is the default behavior." [Source](https://docs.python.org/2/library/subprocess.html#popen-constructor)  

So by my understanding the `env` flag replaces the environment rather than adding it on. And there may be other important variables in the environment already.

clusterfuzz/commands/reproduce.py
everestmz: I agree, I think it would help make some things clearer if there could be a Testcase class. We could then just pass the JSON into the constructor, and have a lot of the methods like `get_testcase_directory` as instance methods instead. Will keep this in mind for another CL.

clusterfuzz/commands/reproduce.py
tanin47: I am more afraid that the inherited env might interfere with reproducing a crash.

On Clusterfuzz, it only use the parsed env.

everestmz: Oh, good point. I'll make this change.

clusterfuzz/commands/reproduce.py
everestmz: I can't split by `=` because the format of these lines is `[Environment] VAR_NAME = var=true:var=false:var=true...` etc. So I need to split it by ` = ` so I only split it into name and value.

tanin47:
Oh I missed that. You're right. It's cool then.

everestmz:
Updated & fixed.

clusterfuzz/commands/reproduce.py
tanin47: I missed that, and you're right.

clusterfuzz/commands/reproduce.py
```diff
@@ -142,7 +145,7 @@ def checkout_chrome_by_sha(sha, chrome_source):
   _, current_sha = common.execute('git rev-parse HEAD',
                                   chrome_source,
                                   print_output=False)
-  if current_sha == sha:
+  if current_sha.strip() == sha:
```
tanin47: When testing this method, we might wanna mock current_sha to have the trailing newline.

PR #11, opened 2017-02-09
Keep v8 source separate from chrome source

everestmz:
This is a small CL, with one motivation: `$CHROME_SRC` may very well be set to point to the actual chrome source code on many people's machines. Since as of now this tool works with V8 testcases only, running on a standalone V8 repo, I think it makes sense to use `$V8_SRC` to avoid needing people to edit their existing environment too much. 

I'm not sure if this is the nicest way to do such a thing, but I think it is useful for this particular case.


clusterfuzz/commands/reproduce.py
```diff
@@ -318,7 +318,16 @@ def execute(testcase_id, current):
 
   response = get_testcase_info(testcase_id)
 
-  chrome_source = os.environ['CHROME_SRC']
+  chrome_source = ''
```
mbarbella: Can we rename this to source_directory or something that's a bit more descriptive?

clusterfuzz/commands/reproduce.py
```diff
@@ -318,7 +318,16 @@ def execute(testcase_id, current):
...
+  if '/v8/v8' in response['regression']:
+    chrome_source = os.environ.get('V8_SRC')
+    if not chrome_source:
+      message = ('This is a V8 testcase, please define $V8_SRC or enter'
+                 ' your V8 source location here: ')
+      chrome_source = os.path.expanduser(raw_input(message))
```
mbarbella: Let's validate this here. I suspect things will go horribly wrong if it's not a valid path, and we might fail in a more confusing way if we don't.

everestmz:
Thanks for the feedback, updated.

clusterfuzz/commands/reproduce.py
```diff
@@ -318,20 +318,31 @@ def execute(testcase_id, current):
...
+        message = 'Please enter a valid directory: '
+  else:
+    source_directory = os.environ['CHROME_SRC']
+
   crash_revision = response['crash_revision']
   testcase_id = response['id']
   if not current:
     git_sha = sha_from_revision(crash_revision)
-    checkout_chrome_by_sha(git_sha, chrome_source)
+    checkout_chrome_by_sha(git_sha, source_directory)
```
mbarbella: I should have looked at this code more closely, but some of these function names will probably need to be updated as well. That can happen in another CL if you'd prefer, but it'd be nice if they were more generic (specifically in that they shouldn't reference chrome, as that we build things other than chrome rather often).

clusterfuzz/commands/reproduce.py
```diff
@@ -318,20 +318,31 @@ def execute(testcase_id, current):
...
   response = get_testcase_info(testcase_id)
 
-  chrome_source = os.environ['CHROME_SRC']
+  source_directory = None
+  if '/v8/v8' in response['regression']:
+    source_directory = os.environ.get('V8_SRC')
+    if not source_directory:
+      message = ('This is a V8 testcase, please define $V8_SRC or enter'
+                 ' your V8 source location here: ')
+      while not (source_directory and os.path.isdir(source_directory)):
```
tanin47: This part of keep asking for an input until it's valid should be in its own method and more generic; All of our ask-for-input should probably use this.

For example, `ask(message, error_message, validate_fn)`. It would be easier to get 100% test coverage as well.

clusterfuzz/commands/reproduce.py
```diff
@@ -318,20 +318,31 @@ def execute(testcase_id, current):
 
   response = get_testcase_info(testcase_id)
 
-  chrome_source = os.environ['CHROME_SRC']
+  source_directory = None
+  if '/v8/v8' in response['regression']:
```
mbarbella: Could you leave a comment explaining why we're doing this and that it's a short-term solution. Seems safe based on our offline discussion, but what's happening here isn't totally straightforward just from the code.

clusterfuzz/commands/reproduce.py
```diff
@@ -318,20 +318,31 @@ def execute(testcase_id, current):
 
   response = get_testcase_info(testcase_id)
 
-  chrome_source = os.environ['CHROME_SRC']
```
mbarbella: See the comment on 337 as well, but it might be worth just using some kind of generic source environment variable. If this is meant to be a short-term solution to just get this working with d8 now I'm fine with it, but it's not ideal long-term.

clusterfuzz/commands/reproduce.py
tanin47: I agree with @mbarbella. This should be a short-term solution specific for d8. When we support more job type, we should probably remove the if-else condition and go with the design pattern where we instantiate an instance based on job type.

One note on this... since we only support d8 job type, we can just assume V8_SRC is there without checking for 'regression'.

clusterfuzz/commands/reproduce.py
```diff
@@ -68,7 +78,8 @@ def get_verification_header():
   """Prompts the user for & returns a verification token."""
 
   webbrowser.open(GOOGLE_OAUTH_URL, new=1, autoraise=True)
-  verification = raw_input('Please enter your verification code:')
+  verification = ask('Please enter your verification code',
+                     'Please enter a code', lambda x: x)
```
everestmz: This is the only other place we ask for raw user input right now, but I think the `ask` method was still a good idea, it will be useful in the future.

everestmz:
Thank you, updated & implemented an `ask` method.

clusterfuzz/commands/reproduce.py
```diff
@@ -47,6 +47,16 @@
         'response_type': 'code',
         'redirect_uri': 'urn:ietf:wg:oauth:2.0:oob'}))
 
+def ask(question, error_message, validate_fn):
```
mbarbella: Not too important at the moment, but if this ends up being used more it may be worth it to define some reasonable default for error_message and validate_fn.

clusterfuzz/commands/reproduce.py
```diff
@@ -318,20 +329,30 @@ def execute(testcase_id, current):
 
   response = get_testcase_info(testcase_id)
 
-  chrome_source = os.environ['CHROME_SRC']
+  #This way of determining source_directory is temporary, as the
```
mbarbella: Nit: spaces after all #s

PR #12, opened 2017-02-10
Clusterfuzz-tools can run a testcase against a downloaded build

everestmz:
cc: @google/clusterfuzz 

I decided to go ahead with this functionality first, as it will help us diagnose which testcases can't be reproduced because we are building incorrectly, or for other reasons.

In addition, the locations in which I'm hardcoding `d8` is increasing. To reduce the issues this could cause later on, one of my next CLs will be to refactor a lot of the testcase-specific logic into a `Testcase` class.


clusterfuzz/commands/reproduce.py
```diff
@@ -312,47 +315,48 @@ def set_up_environment(stacktrace_lines):
...
-    checkout_chrome_by_sha(git_sha, source_directory)
-
   download_build_data(
       response['metadata']['build_url'],
       testcase_id)
-  build_chrome(crash_revision, testcase_id, source_directory)
+  source_directory = get_build_directory(testcase_id)
+
+  if not download:
+    # This way of determining source_directory is temporary, as the
```
everestmz: Moving a block this large into a conditional isn't ideal, but would be undone with the CL that turns testcases into a class, as all of the logic for choosing source directories etc will be abstracted into that class.

clusterfuzz/commands/reproduce.py
```diff
@@ -312,47 +315,48 @@ def set_up_environment(stacktrace_lines):
...
-    git_sha = sha_from_revision(crash_revision)
-    checkout_chrome_by_sha(git_sha, source_directory)
-
   download_build_data(
       response['metadata']['build_url'],
       testcase_id)
-  build_chrome(crash_revision, testcase_id, source_directory)
+  source_directory = get_build_directory(testcase_id)
+
+  if not download:
```
tanin47: This should be a separate method that returns binary directory. 

From the `reproduce()` point of view, it doesn't care if we build locally or use download build. It only wants to know the binary path.

test/clusterfuzz/commands/reproduce_test.py
```diff
@@ -59,7 +59,7 @@ def setUp(self):
   def test_grab_data(self):
     """Ensures all method calls are made correctly."""
     self.mock.sha_from_revision.return_value = '1a2s3d4f'
-    reproduce.execute('1234', False)
+    reproduce.execute('1234', False, False)
```
tanin47: It's pretty strange that our test coverage is 100%... considering that the if condition is never false. There's a bug in our test coverage. Could you investigate the generated coverage html file?

everestmz: Hmm. I actually don't think it's a bug. All the coverage does is makes sure all lines of code are hit, and it does this because we have no `else` block. But yes I should write another case for that test in particular.

tanin47: In clusterfuzz repo, it would report the if condition as not covered (yellow line in the generated html). We might need to configure for it. Let's configure it and add tests to cover that.

You can do this after the testcase class refactoring (because the refactoring might impact the coverage itself).

clusterfuzz/commands/reproduce.py
everestmz: I agree. Do you think it's worth abstracting this right now though? In the next CL that method will be deleted again, and absorbed into the `Testcase` class making everything a lot cleaner.

tanin47: It's your call. This CL is good in itself since the coverage is 100%. You can merge and refactor in the next CL.

everestmz: I'll refactor in the next one to keep the functions separate - this CL enables downloaded builds, the next is a larger refactor.

tanin47: Sound good.

PR #13, opened 2017-02-13
Refactor testcase download & reproduction logic into a Testcase class

everestmz:
cc: @google/clusterfuzz 

I still need to move & refactor tests over, but I would appreciate an initial comment on how this looks - it's a big shift and it would be good to have a quick sanity check to make sure this is going in the correct direction. I think this CL will help keep the code cleaner in the long term.

I'm wondering if this abstracts a bit too much of the download/build/reproduction logic away into `Testcase`, so a second set of eyes would be useful.


clusterfuzz/testcase.py
```diff
@@ -0,0 +1,228 @@
...
+import urlfetch
+import urllib
+import json
+
+from clusterfuzz import common
+
+class TestCase(object):
+  """The TestCase module, to abstract away logic using the testcase JSON."""
+
+  CLUSTERFUZZ_DIR = os.path.expanduser(os.path.join('~', '.clusterfuzz'))
```
tanin47: Move these constants outside the class.

clusterfuzz/commands/reproduce.py
```diff
@@ -126,97 +90,13 @@ def get_testcase_info(testcase_id):
...
-
-
-def sha_from_revision(revision_number):
-  """Converts a chrome revision number to it corresponding git sha."""
-
-  response = urlfetch.fetch(build_revision_to_sha_url(revision_number))
-  return json.loads(response.body)['git_sha']
-
-
 def check_confirm(question):
```
tanin47: This looks like it should be in common.

clusterfuzz/testcase.py
```diff
@@ -0,0 +1,228 @@
...
+  CLUSTERFUZZ_BUILDS_DIR = os.path.join(CLUSTERFUZZ_DIR, 'builds')
+  CLUSTERFUZZ_TESTCASES_DIR = os.path.join(CLUSTERFUZZ_DIR, 'testcases')
+  CLUSTERFUZZ_TESTCASE_URL = ('https://cluster-fuzz.appspot.com/v2/testcase-'
+                              'detail/download-testcase/oauth?id=%s')
+
+  def build_revision_to_sha_url(self):
```
tanin47: I think there should be one more class, Builder (or maybe BinaryProvider). BinaryProvider has two subclasses: V8Builder and DownloadedBinary. They provide `get_binary_path()`. Its constructor can take in the url from a testcase.

V8Builder would build a binary. But DownloadedBinary would do less.

An instance of Testcase should only provide parsed information about a testcase and, maybe, some composed info (e.g. reproduce_args).

The reproduce.py takes these info and reproduce (which is what `reproduce_crash` does).

everestmz: Ok, I agree that makes more sense, this refactor kind of makes `reproduce.py` meaningless which isn't what we wanted. 

Since we call it V8Builder, I guess the general idea is to create a new subclass for each kind of job we can reproduce?

tanin47: Yes, we can refactor V8Builder as we support more job types (e.g. different repos, or some additional build steps). Right now making only one V8Builder is fine.

everestmz:
Thank you for the feedback, updated and refactored further

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,192 @@
...
+  def get_binary_path(self):
+    return '%s/%s' % (self.get_build_directory(), self.target)
+
+  def build_dir_name(self):
+    """Returns a build number's respective directory."""
+    return os.path.join(CLUSTERFUZZ_BUILDS_DIR,
+                        str(self.testcase_id) + '_build')
+
+
+class DownloadedBinary(BinaryProvider):
```
tanin47: We probably need a target name here. Maybe make this class to be V8DownloadedBinary.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,192 @@
...
+  def build_target(self):
+    """Build the correct revision in the source directory."""
+
+    print 'Building revision %i in %s' % (
+        self.revision,
+        self.build_directory)
+
+    self.setup_gn_args()
+    goma_cores = 10 * multiprocessing.cpu_count()
+    #TODO: These two pre-build steps are d8-specific, will need changing.
```
tanin47: Remove? since this class is v8 specific.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,192 @@
...
+
+    self.download_build_data()
+    self.build_directory = self.build_dir_name()
+    return self.build_directory
+
+
+class V8Builder(BinaryProvider):
+  """Builds a fresh v8 binary."""
+
+  def build_revision_to_sha_url(self):
```
tanin47: Move these two methods to above (and outside this class) and parameterise 'v8/v8'.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,192 @@
...
+    return ('https://cr-rev.appspot.com/_ah/api/crrev/v1/get_numbering?%s' %
+            urllib.urlencode({
+                'number': self.revision,
+                'numbering_identifier': 'refs/heads/master',
+                'numbering_type': 'COMMIT_POSITION',
+                'project': 'chromium',
+                'repo': 'v8/v8'})) #TODO: Change this based on testcase JSON
+
+  def sha_from_revision(self):
```
tanin47: Move this method outside as well.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,192 @@
...
+
+  def get_build_directory(self):
+    """Returns the location of the correct build to use for reproduction."""
+
+    if self.build_directory:
+      return self.build_directory
+
+    self.download_build_data()
+    self.build_directory = self.build_dir_name()
+    self.source_directory = os.environ.get('V8_SRC')
```
tanin47: this os.environ.get should be injected through its constructor, so that V8Builder is completely self-contained.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,192 @@
...
+                'project': 'chromium',
+                'repo': 'v8/v8'})) #TODO: Change this based on testcase JSON
+
+  def sha_from_revision(self):
+    """Converts a chrome revision number to it corresponding git sha."""
+
+    response = urlfetch.fetch(self.build_revision_to_sha_url())
+    return json.loads(response.body)['git_sha']
+
+  def __init__(self, testcase, current=False, goma_dir=None):
```
tanin47: I don't think we need to default any argument, since all of them will be set, right?

clusterfuzz/testcase.py
```diff
@@ -0,0 +1,75 @@
...
+    os.makedirs(testcase_dir)
+
+    auth_header = common.get_stored_auth_header()
+    command = 'wget --header="Authorization: %s" "%s" -O ./testcase.js' % (
+        auth_header, CLUSTERFUZZ_TESTCASE_URL % self.id)
+    common.execute(command, testcase_dir)
+
+    return filename
+
+  def setup_environment(self):
```
tanin47: This should be named get_environment(). I think we should just set `self.environments` because it doesn't need external dependencies (as opposed to `get_testcase_path`, where it makes an HTTP request).

clusterfuzz/testcase.py
```diff
@@ -0,0 +1,75 @@
...
+  def testcase_dir_name(self):
+    """Returns a testcases' respective directory."""
+    return os.path.join(CLUSTERFUZZ_TESTCASES_DIR,
+                        str(self.id) + '_testcase')
+
+  def get_testcase_path(self):
+    """Downloads & returns the location of the testcase file."""
+
+    testcase_dir = self.testcase_dir_name()
+    #TODO: Filename testcase.js is d8-specific
```
tanin47: One note: this `Testcase` is actually V8Testcase.  But I think this is fine for this version. We'll refactor it later.

clusterfuzz/testcase.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+
+from clusterfuzz import common
+
+CLUSTERFUZZ_DIR = os.path.expanduser(os.path.join('~', '.clusterfuzz'))
+CLUSTERFUZZ_TESTCASES_DIR = os.path.join(CLUSTERFUZZ_DIR, 'testcases')
+CLUSTERFUZZ_TESTCASE_URL = ('https://cluster-fuzz.appspot.com/v2/testcase-'
+                            'detail/download-testcase/oauth?id=%s')
+
+class TestCase(object):
```
tanin47: Let's do `Testcase` instead of `TestCase`, since clusterfuzz also uses `Testcase`.

clusterfuzz/commands/reproduce.py
```diff
@@ -228,135 +88,30 @@ def ensure_goma():
...
+  else:
+    binary_provider = binary_providers.V8Builder(response, current, goma_dir)
 
-  reproduction_args = get_reproduction_args(response)
-  testcase_file = download_testcase_file(testcase_id)
-  env = set_up_environment(response['crash_stacktrace']['lines'])
-  reproduce_crash(testcase_file, reproduction_args, source_directory, env)
+  binary_location = binary_provider.get_binary_path()
+  current_testcase = testcase.TestCase(response)
+  reproduce_crash(binary_location, current_testcase)
```
tanin47: This is great! No need to use temp variables. We can do:

```
reproduce_crash(
  binary_location = binary_provider.get_binary_path(),
  testcase = testcase.TestCase(response))
```

clusterfuzz/commands/reproduce.py
```diff
@@ -228,135 +88,30 @@ def ensure_goma():
...
-  common.execute('GYP_DEFINES=asan=1 gypfiles/gyp_v8', chrome_source)
-  common.execute(
-      ('ninja -C %s -j %i d8'
-       % (testcase_source_dir, goma_cores)),
-      chrome_source)
-
-
-def get_reproduction_args(testcase_info):
-  """Gets all needed args from testcase info and returns a single string"""
+def reproduce_crash(binary, current_testcase):
```
tanin47: Let's name them `binary_path` and `testcase`.

clusterfuzz/commands/reproduce.py
```diff
@@ -228,135 +88,30 @@ def ensure_goma():
...
-    if not current:
-      git_sha = sha_from_revision(crash_revision)
-      checkout_chrome_by_sha(git_sha, source_directory)
+  goma_dir = ensure_goma()
 
-    build_chrome(crash_revision, testcase_id, source_directory)
-    source_directory = get_out_dir(source_directory, testcase_id)
+  binary_provider = None
+  if download:
+    binary_provider = binary_providers.DownloadedBinary(response)
```
tanin47: I think we can instantiate a testcase first and pass the fields to DownloadedBinary and V8Builder. I notice that it needs only a few fields. It'll be easier to write tests as well because those are strings (as opposed to the whole json). We'll revisit if we need more fields.

Another thing is that testcase should represent all the fields we need. So, it makes sense to instantiate the testcase first and use certain fields to instantiate a BinaryProvider.

clusterfuzz/commands/reproduce.py
everestmz: I use current_testcase because otherwise it interferes with the import of the `testcase` module

tanin47: Well, current_testcase is good, then.

clusterfuzz/commands/reproduce.py
everestmz: There are some parameters (like build url) which aren't specified/needed in  `Testcase` but are needed in the `BinaryProviders` - do you think this still makes sense?

tanin47: Yes, it should be one of parsed fields in Testcase, since Testcase represents parsed info of JSON.

everestmz:
Thank you for the feedback, finally finished refactoring the tests!

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,194 @@
...
+              'numbering_type': 'COMMIT_POSITION',
+              'project': 'chromium',
+              'repo': repo}))
+
+def sha_from_revision(revision, repo):
+  """Converts a chrome revision number to it corresponding git sha."""
+
+  response = urlfetch.fetch(build_revision_to_sha_url(revision, repo))
+  return json.loads(response.body)['git_sha']
+
```
tanin47: Remove extra line.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,194 @@
...
+def build_revision_to_sha_url(revision, repo):
+  return ('https://cr-rev.appspot.com/_ah/api/crrev/v1/get_numbering?%s' %
+          urllib.urlencode({
+              'number': revision,
+              'numbering_identifier': 'refs/heads/master',
+              'numbering_type': 'COMMIT_POSITION',
+              'project': 'chromium',
+              'repo': repo}))
+
```
tanin47: Add a line here.

clusterfuzz/binary_providers.py
```diff
@@ -0,0 +1,194 @@
...
+class BinaryProvider(object):
+  """Downloads/builds and then provides the location of a binary."""
+
+  def __init__(self, testcase_id, build_url):
+    self.testcase_id = testcase_id
+    self.build_url = build_url
+    self.build_directory = None
+    self.target = 'd8'
+
```
tanin47: Add

```
def get_build_directory(self):
   """Get build directory. It must be implemented by a subclass."""
  raise NotImplementedError
```

clusterfuzz/commands/reproduce.py
```diff
@@ -47,60 +37,22 @@
...
-
-
 def get_verification_header():
   """Prompts the user for & returns a verification token."""
 
   webbrowser.open(GOOGLE_OAUTH_URL, new=1, autoraise=True)
-  verification = ask('Please enter your verification code',
-                     'Please enter a code', lambda x: x)
+  verification = common.ask('Please enter your verification code',
+                            'Please enter a code', lambda x: x)
```
tanin47: Use `lambda x: bool(x)`

PR #14, opened 2017-02-15
Use different build directories when checking out a new SHA.

everestmz:
cc @google/clusterfuzz 

This is a small CL, to seperate build (`/out/clusterfuzz_12345...`) directories not only by testcase but by revision too, in an attempt to keep builds as separate as possible. This could help to ensure we have clean directories when reproducing. 

Since we have the `--current` flag we are building two different revisions in the same out folder right now. This may have some effect on the testcase as we aren't always building from a clean folder.


clusterfuzz/binary_providers.py
```diff
@@ -122,17 +122,21 @@ def __init__(self, testcase_id, build_url, revision, current, goma_dir,
...
+    _, current_sha = common.execute('git rev-parse HEAD',
+                                    self.source_directory,
+                                    print_output=False)
+    return current_sha.strip()
+
   def out_dir_name(self):
     return os.path.join(self.source_directory, 'out',
-                        'clusterfuzz_' + str(self.testcase_id))
+                        'clusterfuzz_%s_%s' % (str(self.testcase_id),
+                        self.get_current_sha()[:7]))
```
everestmz: I use the first 7 characters of the Sha. I figure if that's good enough for got to do in some cases, it's good enough for us.

clusterfuzz/binary_providers.py
```diff
@@ -122,17 +122,21 @@ def __init__(self, testcase_id, build_url, revision, current, goma_dir,
     self.revision = revision
     self.git_sha = sha_from_revision(self.revision, 'v8/v8')
 
+  def get_current_sha(self):
+    _, current_sha = common.execute('git rev-parse HEAD',
```
tanin47: Use `git rev-parse --short HEAD` and we don't need to do `:7`.

I see that we need it elsewhere. I actually don't mind having a long folder name though.

clusterfuzz/binary_providers.py
```diff
@@ -122,17 +122,21 @@ def __init__(self, testcase_id, build_url, revision, current, goma_dir,
...
+                                    self.source_directory,
+                                    print_output=False)
+    return current_sha.strip()
+
   def out_dir_name(self):
```
tanin47: Let's add a method description explaining that the dirs are unique based on (sha, is_dirty, testcaseid)

clusterfuzz/binary_providers.py
```diff
@@ -122,17 +122,21 @@ def __init__(self, testcase_id, build_url, revision, current, goma_dir,
...
     return os.path.join(self.source_directory, 'out',
-                        'clusterfuzz_' + str(self.testcase_id))
+                        'clusterfuzz_%s_%s' % (str(self.testcase_id),
```
tanin47: Remove `str()`.

clusterfuzz/binary_providers.py
tanin47: Let's have the long folder name... so that we don't have to add more test for this method.

everestmz:
Thank you, updated!

PR #15, opened 2017-02-16
Add support for reproducing Pdfium Testcases

everestmz:
cc: @google/clusterfuzz 

This CL adds initial support for reproducing Pdfium testcases. There are no tests yet in case we decide to change the architecture a little, but I would appreciate some initial feedback. There are a couple areas that could probably be made more efficient.


clusterfuzz/commands/reproduce.py
```diff
@@ -106,13 +106,24 @@ def execute(testcase_id, current, download):
   response = get_testcase_info(testcase_id)
   goma_dir = ensure_goma()
   current_testcase = testcase.Testcase(response)
+  v8_keywords = ('d8', 'v8')
+  if any(s in current_testcase.job_type for s in v8_keywords):
```
everestmz: It feels like this will get more unwieldy as I add more testcase types -  is there a better way to understand what I'm building beyond parsing through the job type?

clusterfuzz/binary_providers.py
```diff
@@ -211,3 +207,63 @@ def get_build_directory(self):
...
+
+class PdfiumBuilder(GenericBuilder):
+  """Build a fresh Pdfium binary."""
+
+  def get_pdfium_sha(self):
+    """Gets the correct Pdfium sha using the Chromium sha."""
+
+    response = urlfetch.fetch('https://chromium.googlesource.com/chromium/src.git/+/%s/DEPS?format=TEXT' % self.chromium_sha)
+    body = base64.b64decode(response.body)
+    sha_line = [l for l in body.split('\n') if "'pdfium_revision':" in l][0]
```
tanin47: I'm guessing we cannot parse this text as a json string?

clusterfuzz/binary_providers.py
```diff
@@ -211,3 +207,63 @@ def get_build_directory(self):
     self.build_target()
 
     return self.build_directory
+
+
+class PdfiumBuilder(GenericBuilder):
+  """Build a fresh Pdfium binary."""
+
+  def get_pdfium_sha(self):
```
tanin47: Refactor this out to be on the same level with `sha_from_revision`. Let it take the whole URL and return the sha.

clusterfuzz/commands/reproduce.py
tanin47: I guess this is fine for now. When we have more job types, this 'multiplexer' logic needs to be refactor out.

I guess we could even have a dict where it maps a job type to binary_provider.

clusterfuzz/testcase.py
```diff
@@ -60,7 +61,8 @@ def get_testcase_path(self):
 
     testcase_dir = self.testcase_dir_name()
     #TODO: Filename testcase.js is d8-specific
-    filename = os.path.join(testcase_dir, 'testcase.js')
+    file_extension = 'pdf' if 'pdfium' in self.job_type else 'js'
+    filename = os.path.join(testcase_dir, 'testcase.%s' % file_extension)
```
tanin47: I think it might be better to get the extension from `testcase.absolute_path`. Let's make that in a separate CL.

We might need to get the file name from `testcase.absolute_path` as well in the case where the downloded test file is a zip.

everestmz: I agree, this logic is another thing that won't last as we add more testcase types.

clusterfuzz/binary_providers.py
everestmz: Sadly not, it's not really json. It's basically in the format `x = { <json object> }` with a lot of comments, newlines etc around it.

clusterfuzz/binary_providers.py
everestmz: For this method, why would I not just leave the URL inside the method? It won't be used elsewhere.

tanin47: Fair enough. But I think the chromium sha and `pdfium_revision` should be configurable through the arguments.

everestmz:
Tests written, merging now.

PR #18, opened 2017-02-21
Add symbolizing capability to reproduce

everestmz:
cc: @google/clusterfuzz 

This CL enables the use of llvm-symbolizer to symbolize stacktraces of reproduced testcases that use ASAN. As of now, Pdfium tests provide the only symbolize-able stacktrace. For reproducible Pdfium testcases, the symbolizer works properly.


clusterfuzz/commands/reproduce.py
```diff
@@ -91,10 +91,20 @@ def ensure_goma():
 def reproduce_crash(binary_path, current_testcase):
   """Reproduces a crash by running the downloaded testcase against a binary."""
 
+  env = current_testcase.environment
+  #This will need to be dynamic as more jobs are added:
+  symbolizer_variable = 'ASAN_SYMBOLIZER_PATH'
+  if os.environ.get(symbolizer_variable):
```
tanin47: Let's do `if not os.environ.get(symbolizer_variable):` and raise exception.

clusterfuzz/commands/reproduce.py
```diff
@@ -91,10 +91,20 @@ def ensure_goma():
...
+    env[symbolizer_variable] = os.environ[symbolizer_variable]
+  else:
+    raise common.SymbolizerPathError(symbolizer_variable)
+  if 'symbolize=1' not in env['ASAN_OPTIONS']:
+    env['ASAN_OPTIONS'] += ':symbolize=1'
```
tanin47: We don't seem to need this on clusterfuzz. Why do we need it here?

I'm not exactly sure how symbolizer works ... but do we need to pipe output into symbolizer?

everestmz: On clusterfuzz it seems that the default is `symbolize=0`. Could this be because it uses `asan_symbolize.py` instead? Piping isn't needed, all that is required is to provide the location, and then set symbolize to true.

everestmz:
Thank you, suggestions addressed!

PR #19, opened 2017-02-22
Set GOMA options

tanin47:
- Set GOMA options.
- Ignore output because the output of compiling chrome is too large, and it slows down the process.


clusterfuzz/binary_providers.py
```diff
@@ -286,6 +286,7 @@ def __init__(self, testcase_id, build_url, revision, current,
                                           current, goma_dir, source,
                                           binary_name, 'chromium_builder_asan')
     self.git_sha = sha_from_revision(self.revision, 'chromium/src')
+    self.name = 'chromium'
```
everestmz: Why do we need `self.name` here? I don't really see where it's used.

tanin47: It raised an exception because we used it when asking for source_directory.

everestmz: Right, forgot that.

Dor1s:
Do you guys support building without goma? It feels that not everyone is always using it.

tanin47:
Not at the moment. But that's a good point. We'll support it in the future.

PR #24, opened 2017-02-22
Fix Pre-Build steps

everestmz:
cc: @google/clusterfuzz 

This CL makes changes to the steps that generate the args.gn, run pre-build commands such as `gclient runhooks`, and adds extra flags to the `ninja` and `gn gen` commands with the aim of ensuring the build directories are as similar to the remote build as possible.


clusterfuzz/commands/reproduce.py
```diff
@@ -132,16 +132,13 @@ def reproduce_crash(binary_path, current_testcase):
   """Reproduces a crash by running the downloaded testcase against a binary."""
 
   env = current_testcase.environment
-  #This will need to be dynamic as more jobs are added:
-  symbolizer_variable = 'ASAN_SYMBOLIZER_PATH'
-  if not os.environ.get(symbolizer_variable):
-    raise common.SymbolizerPathError(symbolizer_variable)
-
-  env[symbolizer_variable] = os.environ[symbolizer_variable]
+  env['ASAN_SYMBOLIZER_PATH'] = os.path.join(os.path.dirname(binary_path),
```
tanin47: This symbolizer path should be provided by binary_path.

clusterfuzz/binary_providers.py
```diff
@@ -194,6 +197,9 @@ def setup_gn_args(self, other_options=None):
         for k, v in other_options.iteritems():
           f.write('%s = %s\n' % (k, v))
 
+    common.execute('gn gen --check %s' % self.build_directory,
```
tanin47: Let's remove --check since it doesn't work for pdfium.

clusterfuzz/binary_providers.py
```diff
@@ -238,16 +244,15 @@ def __init__(self, testcase_id, build_url, revision, current, goma_dir,
   def setup_gn_args(self, other_options=None):
     other_options = {'pdf_is_standalone': 'true'}
     super(PdfiumBuilder, self).setup_gn_args(other_options)
-    common.execute('gn gen %s' % self.build_directory, self.source_directory)
 
   def build_target(self):
     """Build the correct revision in the source directory."""
 
     self.setup_gn_args()
```
tanin47: We need gclient sync here (but not runhooks).

clusterfuzz/binary_providers.py
everestmz: Since, right now, PDFium is built from the chromium repo, the `--check` is ok. Next CL will add standalone vs chromium capability, and will slightly refactor the build methods to reuse less code. In that, I will ensure that pdfium standalone doesn't run `--check`.

tanin47: sg

PR #27, opened 2017-02-23
Allow selection between downloaded, standalone or chromium build

everestmz:
cc: @google/clusterfuzz 

This allows users to choose between running a testcase against all three possible options.
Putting up for a review before I change tests around.


clusterfuzz/binary_providers.py
```diff
@@ -27,6 +27,8 @@
 
 CLUSTERFUZZ_DIR = os.path.expanduser(os.path.join('~', '.clusterfuzz'))
 CLUSTERFUZZ_BUILDS_DIR = os.path.join(CLUSTERFUZZ_DIR, 'builds')
+CHROMIUM_SUPPORTED_JOBS = {
+    'linux_asan_pdfium': 'pdfium_test'}
```
everestmz: I think it makes sense to support this separately because (as is the case right now) building outside and inside the chromium repo can make a big difference.

clusterfuzz/commands/reproduce.py
```diff
@@ -37,6 +37,12 @@
                       'c7bn50f.apps.googleusercontent.com'),
         'response_type': 'code',
         'redirect_uri': 'urn:ietf:wg:oauth:2.0:oob'}))
+SUPPORTED_JOBS = {
+    'linux_asan_pdfium': [binary_providers.PdfiumBuilder, 'PDFIUM_SRC'],
```
tanin47: Make a class out of this `[binary_providers.PdfiumBuilder, 'PDFIUM_SRC']`.

Let's call it BinaryDefinition. It'll take `binary_providers.PdfiumBuilder` and predefined argument like src and support other generic arguments (stored in a hash).

Then, you can call 

```
definition.builder(
        current_testcase.id, current_testcase.build_url,
        current_testcase.revision, current, goma_dir,
        os.environ.get(definition.source_var), **definition.kwargs)
```

`definition.kwargs` is the hash.

clusterfuzz/binary_providers.py
tanin47: Let's move this definition to SUPPORTED_JOBS.

tanin47: Ok, I see. I understand now. There are two set of job types. One is with standalone repo. Another is with chromium. That makes sense. Let's move this to reproduce. And make it the same format that SUPPORTED_JOBS will use (with BinaryDefinition).

everestmz: The key for `linux_asan_pdfium` is already used in SUPPORTED_JOBS. The issue is, I need some way of passing the binary name to the ChromiumBuilder, since it is generic and doesn't know it by default. What do you think the best way is to achieve that?

everestmz: Ok, sounds good (Y)

everestmz:
Thank you, updated!

clusterfuzz/commands/reproduce.py
```diff
@@ -145,36 +156,40 @@ def reproduce_crash(binary_path, symbolizer_path, current_testcase):
...
+  if build == 'download':
     binary_provider = binary_providers.DownloadedBinary(
         current_testcase.id, current_testcase.build_url, target)
+  elif build == 'standalone':
+    definition = get_binary_definition(current_testcase.job_type)
+    binary_provider = definition.builder( # pylint: disable=redefined-variable-type
+        current_testcase.id, current_testcase.build_url,
+        current_testcase.revision, current, goma_dir,
+        os.environ.get(definition.source_var), **definition.kwargs)
   else:
```
tanin47: Let's do elsif 'chromium' here. And raise exception on else condition.

clusterfuzz/commands/reproduce.py
```diff
@@ -37,6 +37,17 @@
                       'c7bn50f.apps.googleusercontent.com'),
         'response_type': 'code',
         'redirect_uri': 'urn:ietf:wg:oauth:2.0:oob'}))
+SUPPORTED_JOBS = {
+    'linux_asan_pdfium': common.BinaryDefinition(binary_providers.PdfiumBuilder,
```
tanin47: Let's separate this into STANDALONE_SUPPORTED_JOBS and CHROMIUM_SUPPORTED_JOBS.

It's better to keep things separated.

clusterfuzz/commands/reproduce.py
everestmz: It will already error because we use the argument parser, and give it limited options.

tanin47: got it. thanks.

everestmz:
Updated & tests written! I think this needs another approval because I changed some stuff.

clusterfuzz/commands/reproduce.py
```diff
@@ -37,6 +37,21 @@
...
+    'linux_asan_d8': common.BinaryDefinition(binary_providers.V8Builder,
+                                             'V8_SRC', 'd8'),
+    'linux_asan_d8_v8_mipsel_db': common.BinaryDefinition(
+        binary_providers.V8Builder, 'V8_SRC', 'd8'),
+    'linux_v8_d8_tot': common.BinaryDefinition(binary_providers.V8Builder,
+                                               'V8_SRC', 'd8')}
+
+CHROMIUM_SUPPORTED_JOBS = {
+    'linux_asan_pdfium': common.BinaryDefinition(binary_providers.PdfiumBuilder,
+                                                 'PDFIUM_SRC', 'pdfium_test')}
```
tanin47: It looks very good. If I want to add, say, linux_asan_chrome_v8 (no standalone version of this), I can just add it here and call it a day.

PR #31, opened 2017-02-24
Tests for getting binary name and warning for unreproducibles

tanin47:
I missed this.

If we build with a specific target name for fuzzer, it runs much faster.

cc @everestmz


clusterfuzz/common.py
```diff
@@ -25,7 +25,7 @@
 def get_binary_name(stacktrace):
   stacktrace_lines = [l['content'] for l in stacktrace]
   for l in stacktrace_lines:
-    if 'Running command: ' in l:
+    if 'Running command: ' in l or 'Command: ' in l:
```
everestmz: I tried this yesterday - doesn't `Command: ` just refer to running the fuzzer?

tanin47:
Remove the setting target name part. @everestmz will take care of it.

So, I'm submitting what I have left here. They are tests for get_binary_name, warning for unreproducible testcases, and remove GOMA's envs (these are users' responsibility).

PR #33, opened 2017-02-25
Libfuzzer msan support added

everestmz:
cc: @google/clusterfuzz 

- Enables libfuzzer_chrome_msan job type builds/reproduction
- Combines the download and chromium clauses, to reduce branching in `reproduce`
- Changes all `GenericBuilder` subclasses to take Testcase and BinaryDefinition objects rather than many parameters
- moves logic for setting `symbolize=1` into `Testcase`


clusterfuzz/binary_providers.py
```diff
@@ -300,3 +299,13 @@ def out_dir_name(self):
 
   def pre_build_steps(self):
     common.execute('gclient runhooks', self.source_directory)
+
+class LibfuzzerMsanBuilder(ChromiumBuilder):
+  """Builds for a Msan testcase, inside the Chromium repo."""
+
+  def pre_build_steps(self):
+    common.execute(("GYP_DEFINES='clang=1 component=static_library "
+                    "gomadir=/b/build/slave/cache/goma_client msan=1 msan_track"
```
tanin47: Fix the goma path.

clusterfuzz/common.py
```diff
@@ -38,10 +38,12 @@ def get_binary_name(stacktrace):
 class BinaryDefinition(object):
   """Holds all the necessary information to initialize a job's builder."""
 
-  def __init__(self, builder, source_var, binary_name=None, kwargs=None):
+  def __init__(self, builder, source_var, binary_name=None, sanitizer='ASAN',
```
tanin47: Avoid defaulting to 'ASAN'. There's no reason to do so.

PR #34, opened 2017-02-25
Fixed Pdfium parsing binary name incorrectly

### LINKED ISSUE #32: Pdfium doesn't have 'Running command: '
https://github.com/google/clusterfuzz-tools/blob/master/clusterfuzz/commands/reproduce.py#L185

We probably need CI soon to catch this kind of thing.

everestmz:
cc: @google/clusterfuzz 
fixes #32 

Ensures that the binary name is taken from the `BinaryDefinition` if it is present, and also takes args from `window_argument` and `minimized_arguments` if the stacktrace cannot be parsed properly.


clusterfuzz/testcase.py
```diff
@@ -55,13 +55,17 @@ def get_environment_and_args(self):
         l = l[1:len(l)-1] #Strip off the binary & testcase paths
         args = " ".join(l)
 
-    return new_env, args
+    return new_env, None if args == '' else args
```
tanin47: Why do we need to use None instead of ''?

everestmz: I do a check after the method call, using `if not self.reproduction_args` and then set the args the old way if they couldn't be parsed. It doesn't need to be None, I just thought it made it look a little nicer.

PR #36, opened 2017-02-27
Added UBSAN support, updated env parsing

everestmz:
cc: @google/clusterfuzz 

Adds UBSAN support, and also adds more parsing for the ENV to correctly supply suppressions, symbolizer and set LSAN_OPTIONS in cases when LSAN is not the specified sanitizer, but options are still present.


clusterfuzz/commands/reproduce.py
```diff
@@ -139,11 +141,48 @@ def ensure_goma():
   return goma_dir
 
 
+def replace_env_options(options, key, new_value):
```
tanin47: I think it would be better to parse the string into dict (deserialize), modify the dict, and generate the string from dict (serialize). It's similar to how we handle JSON.

clusterfuzz/commands/reproduce.py
```diff
@@ -139,11 +141,48 @@ def ensure_goma():
...
+    return '%s%s=%s' % (options[0:pre_index], key, new_value)
+  return ('%s%s=%s%s' %
+          (options[0:pre_index], key, new_value,
+           options[post_index+pre_index:]))
+
+
+def set_up_symbolizers_suppressions(env, symbolizer_path, sanitizer):
+  """Sets up the symbolizer variables for an environment."""
+
+  parent_folder = os.path.dirname(os.path.dirname(__file__))
```
tanin47: This is going to be problematic when we publish a pip package. But let's worry about it later.

everestmz: I wasn't sure how to best access the files packaged along with the code - I'm sure there is a better way though.

clusterfuzz/commands/reproduce.py
```diff
@@ -139,11 +141,48 @@ def ensure_goma():
...
+  pre_index = options.find(key)
+  post_index = options[pre_index:].find(':')
+  if post_index == -1:
+    return '%s%s=%s' % (options[0:pre_index], key, new_value)
+  return ('%s%s=%s%s' %
+          (options[0:pre_index], key, new_value,
+           options[post_index+pre_index:]))
+
+
+def set_up_symbolizers_suppressions(env, symbolizer_path, sanitizer):
```
tanin47: I wouldn't mind this function to just process all envs that exists.

For example, if MSAN_OPTIONS exists, we modify what we need to modify. We can simply ignore `sanitizer`.

It might be easier to understand what the method does.

everestmz:
Thank you, suggestions addressed :+1:

clusterfuzz/commands/reproduce.py
```diff
@@ -139,11 +141,53 @@ def ensure_goma():
   return goma_dir
 
 
+def deserialize_sanitizer_options(options):
+  """Read options from a variable like ASAN_OPTIONS into a dict."""
+
+  options_split = options.split(':')
```
tanin47: Rename `option_split` to `pairs`.

clusterfuzz/commands/reproduce.py
```diff
@@ -139,11 +141,53 @@ def ensure_goma():
...
+  return_dict = {}
+  for option_pair in options_split:
```
tanin47: Rename `option_pair` to `pair`.

clusterfuzz/commands/reproduce.py
```diff
@@ -139,11 +141,53 @@ def ensure_goma():
...
+  return return_dict
+
+
+def serialize_sanitizer_options(options_dict):
+  """Takes dict of sanitizer options, return a command-line friendly string."""
+
+  return_string = ''
+  for option in options_dict:
+    return_string += ':'
+    return_string += '%s=%s' % (option, options_dict[option])
```
tanin47: Collect `pairs` and use `':'.join(..)`.

```
for key, value in options.iteritems():
  pairs.append('%s=%s' % (key, value))

return ':'.join(pairs)
```

PR #40, opened 2017-02-28
Add stackdriver logging functionality

everestmz:
cc: @tanin47 

I haven't written tests yet, but logging works on both ends. Pushing this for an initial look at the structure.


clusterfuzz/stackdriver_logging.py
```diff
@@ -0,0 +1,79 @@
...
+  """Joins the params dict with info like user id and then sends logs."""
+
+  global session_id
+  session_id = generate_session_id() if not session_id else session_id
+
+  scopes = ['https://www.googleapis.com/auth/logging.write']
+
+  resource_package = __name__
+  resource_path = 'clusterfuzz-tools-logging.json'
+  filename = pkg_resources.resource_filename(resource_package, resource_path)
```
everestmz: I think this pkg_resources way is a nicer way to access static files in a package than the other way we have - let me know what you think.

setup.py
```diff
@@ -28,6 +28,6 @@
             'clusterfuzz = clusterfuzz.main:execute'
         ]
     },
-    install_requires=['urlfetch==1.0.2'],
+    install_requires=['urlfetch==1.0.2', 'httplib2', 'oauth2client'],
```
tanin47: Lock the version for httplib2 and oauth2client.

clusterfuzz/main.py
```diff
@@ -16,6 +16,21 @@
 import argparse
 import importlib
 
+from clusterfuzz import stackdriver_logging
+
+
+def log_exceptions(func):
```
tanin47: Move this method to common.

And it should be used as a decorator:

```
@log
def reproduce(...):
  ...
```

clusterfuzz/main.py
```diff
@@ -16,6 +16,21 @@
...
+  """A decorator that catches all exceptions, logs, and then re-throws."""
+
+  def logged_func(**kwargs):
```
tanin47: called this wrap.

And you need to use `functools.wrap(func)`. Otherwise, `__name__` is not retained. The whole example would be like:

```
def log(func):
  @functools.wrap(func)
  def wrapped(*args, **kwargs):
     try:
       send_start(..)
       func(*args, **kwargs)
     except BaseException as e:
       send_exit(..)
       raise
  return wrapped
```

clusterfuzz/stackdriver_logging.py
```diff
@@ -0,0 +1,79 @@
...
+import json
+import time
+import os
+import binascii
+import pkg_resources
+
+from httplib2 import Http
+from oauth2client.service_account import ServiceAccountCredentials
+
+session_id = None
```
tanin47: I think we should make this the constant `SESSION_ID` and make `get_session_id()` for mocking.

clusterfuzz/stackdriver_logging.py
tanin47: Make the method `read_datafile` in common.

I think we should use data file (https://docs.python.org/3.4/distutils/setupscript.html#installing-additional-files). I'm not exactly sure how to read it. This paragraph might be important:

> Each (directory, files) pair in the sequence specifies the installation directory and the files to install there. If directory is a relative path, it is interpreted relative to the installation prefix (Python’s sys.prefix for pure-Python packages, sys.exec_prefix for packages that contain extension modules). Each file name in files is interpreted relative to the setup.py script at the top of the package source distribution. No directory information from files is used to determine the final location of the installed file; only the name of the file is used.

clusterfuzz/stackdriver_logging.py
```diff
@@ -0,0 +1,79 @@
...
+  return ':'.join([os.environ.get('USER'),
+                   str(time.time()),
+                   str(binascii.b2a_hex(os.urandom(20)))])
+
+
+def send_log(params):
+  """Joins the params dict with info like user id and then sends logs."""
+
+  global session_id
+  session_id = generate_session_id() if not session_id else session_id
```
tanin47: If SESSION_ID is a constant, we can remove these two lines.

clusterfuzz/main.py
```diff
@@ -16,6 +16,21 @@
...
+    try:
+      func(**kwargs)
+    except BaseException as e:
```
tanin47: We need to `send_succeed` after `func(..)` as well.

everestmz: Would that work? I thought execute immediately halts if its command fails (and it will, if the reproduce is successful)

tanin47: I see. Let's find another way to record success.

This seems like we need to parse the stacktrace and compare it with the one in testcase json. Or maybe find each line in crash state in the stacktrace.

But, initially, let's assume that, if we run the whole thing, then it succeeds.

When running a testcase against the build, we can pass `exit_or_error=False` to avoid sys.exit().

everestmz:
Feedback addressed & tests updated and working, thank you!

clusterfuzz/commands/reproduce.py
```diff
@@ -22,6 +22,7 @@
 import urlfetch
 
 from clusterfuzz import common
+from clusterfuzz.stackdriver_logging import log
```
tanin47: import the module (not the method)

PR #57, opened 2017-03-03
Work on chrome ubsan

tanin47:
cc @everestmz 

This change works with non-gesture linux_ubsan_chrome. The env and extra argument are essential.


tanin47:
It's a prototype.

Issue #35, opened 2017-02-27
Access token expires in between getting testcase json and getting testcase file.


tanin47:
We might want to make sure the user can re-login

everestmz:
Fixed by #59

PR #59, opened 2017-03-03
Linux_Ubsan_Chrome builds working

everestmz:
cc: @google/clusterfuzz 

This PR enables reproduction for linux_Ubsan_Chrome builds that don't use gestures, and throws a JobNotSupportedError if they do use gestures. As of now including gestures requires further work.


clusterfuzz/binary_providers.py
```diff
@@ -278,14 +278,19 @@ def pre_build_steps(self):
 class ChromiumBuilder(GenericBuilder):
   """Builds a specific target from inside a Chromium source repository."""
 
-  def __init__(self, testcase, binary_definition, current, goma_dir):
+  def __init__(self, testcase, binary_definition, current, goma_dir,
+               target=None):
```
tanin47: The target seems redundant here. Below we use target name.

clusterfuzz/commands/reproduce.py
```diff
@@ -61,7 +61,11 @@
             sanitizer='MSAN'),
         'libfuzzer_chrome_ubsan': common.BinaryDefinition(
             binary_providers.ChromiumBuilder, 'CHROMIUM_SRC',
-            sanitizer='UBSAN')}}
+            sanitizer='UBSAN'),
+        'linux_ubsan_chrome': common.BinaryDefinition(
+            binary_providers.ChromiumBuilder, 'CHROMIUM_SRC', 'chrome',
+            sanitizer='UBSAN', target='chromium_builder_asan',
+            extra_runtime_args='--user-data-dir=/tmp/user-profile-data')}}
```
tanin47: This is special folder, and we need to nuke the folder every time. I don't mind doing in the testcase on every run.

This seems like we need the third class: Reproducer. (We have Testcase which represents a testcase. We have Builder which represents how we build.)

The reproduce_crash function is basically Reproducer class.

everestmz: Ah, I didn't know that. Reproducer sounds like a good idea then.

everestmz:
Thank you, feedback addressed!

everestmz:
also fixes #35

clusterfuzz/reproducers.py
```diff
@@ -0,0 +1,96 @@
...
+                   environment=self.environment, exit_on_error=False)
+
+
+class LinuxUbsanChromeReproducer(BaseReproducer):
+  """Adds and extre pre-build step to BaseReproducer."""
+
+  def pre_build_steps(self):
+    """Steps to run before building."""
+
+    user_profile_dir = '/tmp/user-profile-data'
```
tanin47: Let's throw in the word 'clusterfuzz' into the folder name.

clusterfuzz/reproducers.py
```diff
@@ -0,0 +1,96 @@
...
+    if os.path.exists(user_profile_dir):
+      shutil.rmtree(user_profile_dir)
+    self.args += ' --user-data-dir=/tmp/user-profile-data'
```
tanin47: Use the variable user_profile_dir instead of string.

clusterfuzz/binary_providers.py
```diff
@@ -280,12 +280,16 @@ class ChromiumBuilder(GenericBuilder):
 
   def __init__(self, testcase, binary_definition, current, goma_dir):
 
+    target_name = None
     binary_name = binary_definition.binary_name
+    if hasattr(binary_definition, 'target'):
```
tanin47: I start to regret the **kwargs design decision.

I think we should explicitly define `target` and the field we need in BinaryDefinition.

everestmz: I agree, it makes more sense to be explicit.

PR #60, opened 2017-03-06
linux_msan_chrome testcases with no gestures working

everestmz:
cc: @google/clusterfuzz 

This enables linux_msan_chrome jobs with no gestures. The next step is to implement gesture support for MSAN and UBSAN job types.


clusterfuzz/commands/reproduce.py
```diff
@@ -71,7 +71,11 @@
         'linux_ubsan_chrome': common.BinaryDefinition(
             binary_providers.ChromiumBuilder, 'CHROMIUM_SRC',
             reproducers.LinuxUbsanChromeReproducer, 'chrome',
-            sanitizer='UBSAN', target='chromium_builder_asan')}}
+            sanitizer='UBSAN', target='chromium_builder_asan'),
+        'linux_msan_chrome': common.BinaryDefinition(
+            binary_providers.ChromiumBuilder, 'CHROMIUM_SRC',
+            reproducers.LinuxUbsanChromeReproducer, 'chrome',
```
everestmz: The same build steps that work for UBSAN work for MSAN. If it works for other things as well it may be worth just changing this into LinuxChromeBuilder or something similar.

tanin47: Renaming it would be great.

PR #62, opened 2017-03-09
Gestures work for linux_MSAN and linux_UBSAN chrome job types

everestmz:
cc: @google/clusterfuzz 

This PR adds gesture support to to the chrome jobs, meaning all msan and ubsan chrome testcases should now be reproducible.


clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+
+  # Try to find the running process.
+  if not psutil.pid_exists(process_id):
+    return []
+
+  pids = [process_id]
+  try:
+    psutil_handle = psutil.Process(process_id)
+    children = (psutil_handle.children(recursive=recursive) if
+                psutil.version_info >= (2, 0) else
```
tanin47: Remove this, and choose one of them because we lock the version of psutil.

clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+
+def find_windows_for_process(process_id, display_name):
+  """Return visible windows belonging to a process."""
+  pids = get_process_ids(process_id)
+  if not pids:
+    return []
+
+  visible_windows = []
+  for _ in pids:
+    windows = ''
```
tanin47: Remove.

clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+    _, windows = execute(('xdotool search --all --pid %s --onlyvisible --nam'
+                          'e ".*"' % display_name), os.path.expanduser('~'),
+                         environment={'DISPLAY': display_name},
+                         exit_on_error=False)
+    for line in windows.splitlines():
+      if not line.isdigit():
+        continue
+      if int(line) not in visible_windows:
```
tanin47: Change `visible_windows` to set. And use `visible_windows.add(line)`. If we don't need to convert it to int(), let's not do it (shorter code).

clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+def execute(command,
+            cwd,
+            print_output=True,
+            capture_output=True,
+            exit_on_error=True,
+            environment=None,
+            gestures=None):
+  """Execute a bash command."""
+  print 'Running: %s' % command
+  if gestures:
```
tanin47: I don't think we should combine gestures into this method. Let's move gestures' logic to reproducer.

And we don't need to do if gestures. Let's just call a different method for it.

clusterfuzz/testcase.py
```diff
@@ -72,6 +72,10 @@ def __init__(self, testcase_json):
     self.file_extension = self.get_file_extension(
         testcase_json['testcase']['absolute_path'])
     self.reproducible = not testcase_json['testcase']['one_time_crasher_flag']
+    if 'gestures' in testcase_json['testcase']:
```
tanin47: `self.gestures = testcase_json['testcase'].get('gestures')`

clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+        break
+    environment['DISPLAY'] = display_name
+    blackbox = subprocess.Popen(['blackbox'], env={'DISPLAY': display_name})
+    print "blackbox"
+    time.sleep(30)
+
+  proc = start_execute(command, cwd, environment)
+
+  if gestures:
+    if gestures[-1].startswith('Trigger'):
```
tanin47: This `if-else` could be a separate method. It would be easier to test.

clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+    for window in windows:
+      print "Window %i of %i" % (i, len(windows))
+      i += 1
+      xdotool_command('windowactivate --sync %d' % window, display_name)
+
+      for gesture in gestures:
+        print gesture
+
+        gesture_type, gesture_cmd = gesture.split(',')
+        if gesture_type == 'windowsize':
```
tanin47: This `if-else` could be a separate method. It would be easier to get 100% test coverage.

clusterfuzz/common.py
```diff
@@ -175,6 +176,113 @@ def _print(s):
...
+    if gestures[-1].startswith('Trigger'):
+      gesture_start_time = int(gestures[-1].split(':')[1])
+      gestures.pop()
+    else:
+      gesture_start_time = DEFAULT_GESTURE_TIME
+    time.sleep(gesture_start_time)
+
+    windows = find_windows_for_process(proc.pid, display_name)
+    i = 0
+    for window in windows:
```
tanin47: Use `for index, window in enumerate(windows)` and remove the need for `i`.

everestmz:
Thank you, suggestions addressed & tests updated!

clusterfuzz/reproducers.py
```diff
@@ -83,9 +90,112 @@ def reproduce_crash(self):
...
+    else:
+      self.xdotool_command('%s -- %s' % (gesture_type, gesture_cmd),
+                           display_name)
+
+  def get_gesture_start_time(self):
+    """Determine how long to sleep before running gestures."""
+
+    if self.gestures[-1].startswith('Trigger'):
+      gesture_start_time = int(self.gestures[-1].split(':')[1])
+      self.gestures.pop()
```
tanin47: This has a side effect on `self.gestures`. We want to avoid it in general.

I think we should parse `testcase.gestures` into `self.gesture_start_time` and `self.gestures` in the very beginning. And we can just use those variables.

clusterfuzz/reproducers.py
```diff
@@ -28,6 +34,7 @@ def __init__(self, binary_provider, testcase, sanitizer):
     self.binary_path = binary_provider.get_binary_path()
     self.symbolizer_path = binary_provider.symbolizer_path
     self.sanitizer = sanitizer
+    self.gestures = testcase.gestures
```
tanin47: Parse `testcase.gestures` into `self.gestures` and `self.gesture_start_time` here.

clusterfuzz/reproducers.py
```diff
@@ -83,9 +102,102 @@ def reproduce_crash(self):
...
+    """Return visible windows belonging to a process."""
+    pids = self.get_process_ids(process_id)
+    if not pids:
+      return []
+
+    visible_windows = set()
+    for _ in pids:
+      _, windows = common.execute(
+          ('xdotool search --all --pid %s --onlyvisible --nam'
+           'e ".*"' % display_name), os.path.expanduser('~'),
```
tanin47: We are not using `pid` here. It's strange that it even returns window numbers.

Issue #48, opened 2017-03-02
Catch exceptions and print error messages without stacktraces


Dor1s:
For example, an attempt to use some unsupported job causes an exception:

```
$ clusterfuzz reproduce 6263517746561024 -b chromium
<...>
Traceback (most recent call last):
  File "/usr/local/bin/clusterfuzz", line 11, in <module>
    sys.exit(execute())
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/main.py", line 42, in execute
    command.execute(**arg_dict)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/stackdriver_logging.py", line 89, in wrapped
    func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/commands/reproduce.py", line 231, in execute
    definition = get_binary_definition(current_testcase.job_type, build)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/commands/reproduce.py", line 206, in get_binary_definition
    raise common.JobTypeNotSupportedError(job_type)
clusterfuzz.common.JobTypeNotSupportedError: The job libfuzzer_chrome_asan_debug is not yet supported by clusterfuzz tools.
```

Can we catch it and just print error message without the stacktrace?

tanin47:
I think we can have the `ExpectedException` type where we print the error message without printing the whole stacktrace.

PR #63, opened 2017-03-10
Don't show stacktrace for Clusterfuzz exceptions

everestmz:
cc: @google/clusterfuzz 

A small PR to make expected exceptions look nicer when they error out.
Fixes: #48


clusterfuzz/stackdriver_logging.py
```diff
@@ -105,5 +106,8 @@ def wrapped(*args, **kwargs):
     except BaseException as e:
       send_failure(e.__class__.__name__, *args,
                    command=command_name, **kwargs)
+      if issubclass(e.__class__, common.ClusterfuzzException):
```
tanin47: Use another except clause above `except BaseException` instead of `if issubclass`.

clusterfuzz/common.py
```diff
@@ -36,7 +36,14 @@ def get_binary_name(stacktrace):
                   prefix)
 
 
-class SanitizerNotProvidedError(Exception):
+class ClusterfuzzException(Exception):
```
tanin47: I think we should name it `ExpectedException`. Because the exception is expected and will not print a stacktrace. On another hand, a normal exception will print a stacktrace.

PR #66, opened 2017-03-10
Added option to disable GOMA

### LINKED ISSUE #54: An option to enable/disable goma.
By default, we will inject `is_goma=true` if the flag is not present in `args.gn`.  Strangely, buildbot already set `goma_dir` but not `is_goma`.

We can make it nicer by detecting if goma is installed and ensure both `is_goma` and `goma_dir`.

As the first step, let's just inject `is_goma=true` into `args.gn`.

everestmz:
cc: @google/clusterfuzz 
Fixes #54 by adding a `--disable-goma` flag which removes goma details from args.gn.


clusterfuzz/binary_providers.py
```diff
@@ -195,6 +195,10 @@ def setup_gn_args(self):
       for line in lines:
         if 'goma_dir' in line:
           line = 'goma_dir = "%s"' % self.goma_dir
+          if not self.goma_dir:
```
tanin47: I think this is warrant for parsing args.gn properly into a dict, change the values, and serialize it back.

clusterfuzz/binary_providers.py
```diff
@@ -216,7 +220,8 @@ def build_target(self):
     common.execute('gclient sync', self.source_directory)
     #Note: gclient sync must be run before setting up the gn args
     self.setup_gn_args()
-    goma_cores = 10 * multiprocessing.cpu_count()
+    cpu_count = multiprocessing.cpu_count()
+    goma_cores = 10 * cpu_count if self.goma_dir else 3/4 * cpu_count
```
tanin47: Move this into its own method, so it's easier to achieve 100% test coverage.

everestmz:
Thank you for the feedback, updated!

PR #68, opened 2017-03-10
Fix response condition

tanin47:
Fun fact: 

The `Response` class of `requests` overrides `__nonzero__` (https://github.com/kennethreitz/requests/blob/d6f4818c0b40bc6c00433c013b7daaea83b2cd51/requests/models.py#L671).

This means that, if the status is 400, `bool(response)` becomes False.


tanin47:
It's because the switch from `urlfetch` to `requests`.

PR #70, opened 2017-03-10
Only run gestures when they exist

everestmz:
cc: @google/clusterfuzz 
A quick PR to guard running gestures to ensure the extra commands aren't always run.


test/clusterfuzz/reproducers_test.py
```diff
@@ -157,6 +157,7 @@ def test_reproduce_crash(self):
 
     reproducer = reproducers.LinuxChromeJobReproducer(
         mocked_provider, mocked_testcase, 'UBSAN')
+    reproducer.gestures = True
```
tanin47: Should we use a list of string instead of `True`?

everestmz: We can. It doesn't matter here because all it does is trigger calling the run_gestures method, but I'll change it.

PR #72, opened 2017-03-13
Use pid instead of display_name to find chrome windows

everestmz:
cc: @google/clusterfuzz 
A small PR to ensure we are finding chrome windows the correct way.


clusterfuzz/reproducers.py
```diff
@@ -160,11 +160,12 @@ def find_windows_for_process(self, process_id, display_name):
     if not pids:
       return []
 
+    time.sleep(20)
     visible_windows = set()
-    for _ in pids:
+    for process in pids:
```
tanin47: name it `pid`

clusterfuzz/reproducers.py
```diff
@@ -160,11 +160,12 @@ def find_windows_for_process(self, process_id, display_name):
...
       _, windows = common.execute(
           ('xdotool search --all --pid %s --onlyvisible --nam'
```
tanin47: I just noticed this. Could you put `name` on the same line and wrap semantically instead?

PR #81, opened 2017-03-14
Enabled debug text suppression and logging

everestmz:
cc: @google/clusterfuzz 
Most `print` statements are now either `logger.debug` or `logger.info` statements, apart from ones that print nothing and are just present for nice spacing. These don't need to be in the log file.

Suppresses large amounts of output (i.e. output from ninja commands) and logs them to a file only, unless the command is run with `CF_DEBUG=1`, in which case everything is printed as usual.


clusterfuzz/reproducers.py
```diff
@@ -98,9 +100,18 @@ def reproduce_crash(self):
     self.pre_build_steps()
 
     command = '%s %s %s' % (self.binary_path, self.args, self.testcase_path)
-    common.execute(command, os.path.dirname(self.binary_path),
-                   environment=self.environment, exit_on_error=False)
+    return common.execute(command, os.path.dirname(self.binary_path),
+                          environment=self.environment, exit_on_error=False)
 
+  def reproduce(self):
```
everestmz: I think it makes sense to have this method, as this CL ensures that output from `execute` is suppressed unless specified otherwise. Since the idea is to take the stacktrace, check if it is correct, and then print it out, it is better to just print it right here, as opposed to having a temporary workaround to still print from within `execute`.

In a later CL, I can then use this method to check the validity of the stacktrace first.

clusterfuzz/commands/reproduce.py
```diff
@@ -199,8 +200,8 @@ def maybe_warn_unreproducible(current_testcase):
   """Print warning if the testcase is unreproducible."""
   if not current_testcase.reproducible:
     print
-    print ('WARNING: The testcase %s is marked as unreproducible. Therefore,'
-           ' it might not be reproduced correctly here.')
+    logger.info(('WARNING: The testcase %s is marked as unreproducible. '
```
mbarbella: I can see an argument for using info here, but is there any specific reason not to just use warning and change the message?

clusterfuzz/commands/reproduce.py
```diff
@@ -209,16 +210,17 @@ def maybe_warn_unreproducible(current_testcase):
...
+               testcase_id, current, build, disable_goma)
+  logger.info('Downloading testcase information...')
 
   response = get_testcase_info(testcase_id)
   current_testcase = testcase.Testcase(response)
 
   if 'gestures' in response['testcase']:
-    print ('Warning: testcases using gestures are still in development and are '
-           'not guaranteed to reproduce correctly.')
+    logger.info(('Warning: testcases using gestures are still in development '
```
mbarbella: Ditto.

clusterfuzz/common.py
```diff
@@ -162,17 +166,22 @@ def wait_execute(proc, exit_on_error, capture_output=True, print_output=True):
 
   def _print(s):
     if print_output:
-      print s
+      logger.info(s)
 
+  _print('---------------------------------------')
```
mbarbella: Nit: if this was intended to be a specific length, '-' * length is a bit cleaner. If this was arbitrary and just looks right at this size it seems fine as-is.

everestmz: This is arbitrary, I was just going off one of Abhishek's previous suggestions to start & end command output with a barrier like that. Not sure if this is the best way to do it, but it's pretty clear.

PR #83, opened 2017-03-16
Introduce timeout, and check if stacktrace is correct

everestmz:
Uses the clusterfuzz api to get information about a stacktrace, and compares that with the original information. Also times out testcases that don't crash themselves after 15 seconds.

An initial push for some comments on the code. Since we can't do `Popen.wait(timeout=..` (python 3.x feature) I think this is the next best method.

cc: @google/clusterfuzz


clusterfuzz/common.py
```diff
@@ -158,10 +160,11 @@ def start_execute(command, cwd, environment):
       stdout=subprocess.PIPE,
       stderr=subprocess.STDOUT,
       cwd=cwd,
-      env=environment)
+      env=environment,
+      preexec_fn=os.setsid)
```
tanin47: Remove

clusterfuzz/reproducers.py
```diff
@@ -109,9 +114,16 @@ def reproduce(self):
     _, output = self.reproduce_crash()
 
     print
```
tanin47: Use a logger instead?

clusterfuzz/common.py
everestmz: I think the preexec_fn is necessary - it basically sets a group ID so that later when I do `os.killpg` it kills the entire process group including all children.

clusterfuzz/reproducers.py
everestmz: Since it's just for spacing, it will print unnecessary newlines in the logger which isn't all that useful and just takes up space

clusterfuzz/common.py
tanin47: Where is it used? I can't see it.

tanin47: Oh I see. I thought it was an declared argument. It's good then.

everestmz: https://github.com/google/clusterfuzz-tools/pull/83/files#diff-eeeedb3bfb8a09fb039382ee9fc5271bR182 Here, to kill the process if it doesn't stop on its own after the polling interval

everestmz: Ok, cool :+1:

PR #84, opened 2017-03-16
Attempt to reproduce a testcase continuously

everestmz:
cc: @google/clusterfuzz 
Since we now know whether or not a testcase was reproduced correctly, this CL adds a looping mechanism so that testcases are constantly run until either a reproduction occurs, or the user exits the program.


clusterfuzz/reproducers.py
```diff
@@ -111,25 +111,35 @@ def reproduce_crash(self):
...
+          logger.info(output)
+          logger.info('The stacktrace matches the original crash')
+          return True
+        else:
+          logger.info('Reproduction attempt %d unsuccessful. Press Ctrl+C to'
+                      ' stop trying to reproduce.', iterations)
+          iterations += 1
+          logger.debug(output)
+    except KeyboardInterrupt:
+      logger.info('Most recent stacktrace: ')
```
tanin47: I'm inclined to print the stacktrace every time it doesn't reproduce. Showing only the latest stacktrace could be misleading.

clusterfuzz/stackdriver_logging.py
```diff
@@ -111,7 +111,7 @@ def wrapped(*args, **kwargs):
         send_failure(e.__class__.__name__, *args,
                      command=command_name, **kwargs)
         raise
-    except common.ExpectedException as e:
+    except (KeyboardInterrupt, common.ExpectedException) as e:
```
tanin47: Good call.

test/clusterfuzz/reproducers_test.py
```diff
@@ -426,14 +426,21 @@ def test_good_stacktrace(self):
     self.assert_exact_calls(self.mock.reproduce_crash, [
         mock.call(self.reproducer)])
 
+  def _raise(self, *unused_args):
+    raise KeyboardInterrupt
+
   def test_bad_stacktrace(self):
     """Tests functionality when the stacktraces don't match."""
+    helpers.patch(self, ['logging.Logger.debug', 'logging.Logger.info'])
+    self.mock.debug.side_effect = self._raise
```
tanin47: `self.mock.debug.side_effect = KeyboardInterrupt()` is equivalent.

Also, let's run it a few times before raising an exception. Let's do:

`self.mock.debug.side_effect = [None, None, KeyboardInterrupt()]`

clusterfuzz/reproducers.py
everestmz: I agree. I did this to keep console spam to a minimum but this is really the one place where more is better.

everestmz:
Thank you, updated with suggestions

clusterfuzz/reproducers.py
```diff
@@ -111,25 +111,34 @@ def reproduce_crash(self):
...
+        response = requests.post(
+            url=('https://staging-dot-cluster-fuzz.appspot.com/v2/'
+                 'parse_stacktrace'),
+            data=json.dumps({'job': self.job_type, 'stacktrace': output}))
+        response = json.loads(response.text)
+        new_crash_state = [x for x in response['crash_state'].split('\n') if x]
+
+        if (new_crash_state == self.crash_state and
+            response['crash_type'] == self.crash_type):
+          logger.info(output)
```
tanin47: Move this outside of if and remove the below `logging.info(output)`.

clusterfuzz/reproducers.py
```diff
@@ -111,25 +111,34 @@ def reproduce_crash(self):
...
+          logger.info('The stacktrace matches the original crash')
+          return True
+        logger.info(output)
+        logger.info('Reproduction attempt %d unsuccessful. Press Ctrl+C to'
+                    ' stop trying to reproduce.', iterations)
+        iterations += 1
+    except KeyboardInterrupt:
+      logger.info('Most recent stacktrace: ')
```
tanin47: If we print output every time, we don't need this `try-catch` anymore.

PR #86, opened 2017-03-16
Add human-readable message to stackdriver logs

everestmz:
cc: @google/clusterfuzz 
Logs are now sent with a `message` field that looks like:
`everestmz unsuccessfully finished running reproduce with testcase=6061368416665600, build_type=chromium, current=False, and goma enabled` 
`everestmz started running reproduce with testcase=6061368416665600, build_type=chromium, current=False, and goma enabled`


clusterfuzz/stackdriver_logging.py
```diff
@@ -49,6 +49,18 @@ def send_log(params):
 
   params['user'] = os.environ.get('USER')
   params['sessionId'] = get_session_id()
+  if 'success' in params:
+    prefix = ('successfully finished' if params['success'] else
+              'unsuccessfully finished')
+  else:
+    prefix = 'started'
+  params['message'] = ('%s %s running %s with testcase=%s, build_type=%s, '
+                       'current=%s, and goma %s' % (
```
tanin47: add = after goma to be consistent.

PR #90, opened 2017-03-20
Symbolize chrome builds via a llvm-symbolizer proxy

everestmz:
@tanin47


everestmz:
Working & tested!

clusterfuzz/asan_symbolize_proxy.py
```diff
@@ -0,0 +1,24 @@
...
+  resource_package = __name__
+  llvm_location = pkg_resources.resource_filename(resource_package,
+                                                  'llvm-symbolizer')
+  cmd = [llvm_location]
+  for x in argv[1:]:
+    if '--functions' in x:
+      cmd.append('--functions=linkage')
+    else:
+      cmd.append(x)
+  subprocess.Popen(cmd, stdin=sys.stdin, stdout=sys.stdout)
```
tanin47: Can we use call or something alike? Popen is not a blocking call.

clusterfuzz/reproducers.py
```diff
@@ -244,9 +244,27 @@ def pre_build_steps(self):
...
     super(LinuxChromeJobReproducer, self).pre_build_steps()
 
+
+  def post_run_symbolize(self, output):
+    """Symbolizes non-libfuzzer chrome jobs."""
+
+    asan_symbolizer_location = os.path.join(
+        os.environ['CHROMIUM_SRC'], os.path.join('tools', 'valgrind', 'asan',
+                                                 'asan_symbolize.py'))
+    command = 'echo "%s" | %s' % (output, asan_symbolizer_location)
```
tanin47: This is a bash way to sending string to stdin. Can we do this instead? http://stackoverflow.com/a/8475367/710895

PR #95, opened 2017-03-21
Parse both stacktraces using the API to avoid discrepancies.

everestmz:
cc: @google/clusterfuzz


clusterfuzz/reproducers.py
```diff
@@ -108,6 +111,18 @@ def reproduce_crash(self):
     return common.execute(command, os.path.dirname(self.binary_path),
                           environment=self.environment, exit_on_error=False)
 
+  def get_stacktrace_info(self, trace):
+    """Post a stacktrace, return (crash_state, crash_type)."""
+
+    response = requests.post(
+        url=('https://staging-dot-cluster-fuzz.appspot.com/v2/'
```
tanin47: Use the prod one.

PR #96, opened 2017-03-22
Only ask for source directory at the start of each run

everestmz:
cc: @google/clusterfuzz 
Uses the source directory from binary_provider so we don't need to ask for it later on.


tanin47:
Approved. Let's merge.

PR #99, opened 2017-03-22
Move tool to subdirectory and make a CI directory

everestmz:
cc: @google/clusterfuzz


tool/bootstrap.py
```diff
@@ -4,14 +4,21 @@
...
+  return pkg_resources.resource_filename(LOCATION, filename)
 
 def after_install(_, home_dir):
   """Perform additional steps after installation."""
   pip_path = os.path.join(home_dir, 'bin', 'pip')
+  location = __name__
   subprocess.call([pip_path, 'install', '-U', 'pip'])
-  subprocess.call([pip_path, 'install', '-r', 'requirements.txt'])
-  subprocess.call([pip_path, 'install', '-e', '.'])
+  subprocess.call([pip_path, 'install', '-r', get_path('requirements.txt')])
```
tanin47: Can we do `tool/requirement.txt` instead?

We want to force user to run from the root dir anyway.

everestmz: Oh yeah, if we want to force them to then yes

PR #106, opened 2017-03-23
Don't use goma if use_goma is not present

everestmz:
cc: @google/clusterfuzz 
Fixes: #104 

We had the case when the user specified not to use goma, and when `use_goma=false`, but not when `use_goma` isn't present. This fixes that.


clusterfuzz/binary_providers.py
```diff
@@ -193,8 +193,9 @@ def serialize_gn_args(self, args_hash):
   def setup_gn_goma_params(self, gn_args):
     """Ensures that goma_dir and gn_goma are used correctly."""
 
-    if not self.goma_dir or (
-        'use_goma' in gn_args and gn_args['use_goma'] == 'false'):
+    if (not self.goma_dir or
+        ('use_goma' in gn_args and gn_args['use_goma'] == 'false') or
```
tanin47: `gn_args.get('use_goma', 'false') == 'false'` is shorter.

clusterfuzz/binary_providers.py
```diff
@@ -240,6 +241,7 @@ def get_goma_cores(self):
     """Choose the correct amount of GOMA cores for a build."""
 
     cpu_count = multiprocessing.cpu_count()
+    import pdb; pdb.set_trace()
```
tanin47: remove

PR #107, opened 2017-03-23
Ensure ninja progress bar resizes with console

everestmz:
cc: @google/clusterfuzz 
Fixes: #101 
`get_terminal_size` is only available in python 3.x so we had to use `backports.shutil_get_terminal_size`.


everestmz:
I rebuilt on CircleCI without cache, and I think that circle was caching all of the packages, because the build failed due to bootstrap not working (for the same issue we had before - CircleCI already has a virtualenv). I changed circle.yml to run the original commands again, and it works perfectly.

PR #110, opened 2017-03-28
Allow manual choice of goma job number

everestmz:
cc: @google/clusterfuzz 
Allows user to specify a custom number of goma cores by adding `-j <number>` as an arg.


tool/clusterfuzz/main.py
```diff
@@ -37,6 +37,9 @@ def execute(argv=None):
   reproduce.add_argument(
       '-dg', '--disable-goma', action='store_true', default=False,
       help='Disable GOMA when building binaries locally')
+  reproduce.add_argument(
+      '-j', action='store', default=None,
```
tanin47: Add type=int https://docs.python.org/2/library/argparse.html#example

tool/test/clusterfuzz/main_test.py
```diff
@@ -32,20 +32,20 @@ def setUp(self):
   def test_parse_reproduce(self):
     """Test parse reproduce command."""
     main.execute(['reproduce', '1234'])
-    main.execute(['reproduce', '1234', '--current'])
+    main.execute(['reproduce', '1234', '-j 25' ,'--current'])
```
tanin47: Use '-j' and '25'

PR #111, opened 2017-03-29
Add ability to disable gclient sync and runhooks

everestmz:
cc: @google/clusterfuzz 
Allows the user to include `--disable-gclient-commands` which cancels all runhooks and sync commands from being run.


tool/clusterfuzz/common.py
```diff
@@ -258,6 +258,8 @@ def execute(command, cwd, print_output=True, capture_output=True,
             exit_on_error=True, environment=None):
   """Execute a bash command."""
 
+  if os.environ.get('DISABLE_GCLIENT_COMMANDS') and 'gclient' in command:
```
tanin47: I think we should inject `disable_gclient_commands` to all builders instead and builders decide not to run gclient.

Plus, common.execute() shouldn't have knowledge about the layer above it (which is the business logic).

everestmz: I would agree, it just feels like the builders are receiving too many arguments that do just one thing. And in this specific case, the pre build commands may have very different commands for each builder, which would basically mean adding some if statements to each builder's implementation of `pre_build_steps`. I do however agree that `execute` really shouldn't know a lot.

tanin47: Builders having a large responsibility is ok in this case because this one is a builder decision not to do `gclient`. If the responsibility gets too large, we can refactor the builder in that case.

everestmz:
Thank you, updated to modify binary_providers.

PR #114, opened 2017-03-29
Add quiet mode for running on CI

everestmz:
cc: @google/clusterfuzz


tool/clusterfuzz/common.py
```diff
@@ -269,6 +269,9 @@ def confirm(question, default='y'):
   default can either be 'y', 'n', or None. Answer
   is returned as either True or False."""
 
+  if os.environ.get('CF_QUIET'):
+    return True if default == 'y' else False
```
tanin47: Why not just return true? Because, when CF_QUIET, we want to go through every confirm, right?

Returning false will generally mean exit.

everestmz: Yeah, that does make more sense. Good point

PR #120, opened 2017-03-30
Release v0.2.2

everestmz:
cc: @google/clusterfuzz 
Release 0.2.2 with support for choosing the number of iterations so CI can be tested with release builds.


tanin47:
Make it an RC, and let's lock the version on co first

tanin47:
We need to test before publishing a new version. It's a bit risky to
publish again now.

PR #135, opened 2017-04-03
Set default max iterations to 10

### LINKED ISSUE #125: Let's default the iteration to 10
Making it run infinitely is a bit too extreme.

everestmz:
cc: @google/clusterfuzz 
Fixes #125


tool/clusterfuzz/main.py
```diff
@@ -44,7 +44,7 @@ def execute(argv=None):
       '--disable-gclient-commands', action='store_true', default=False,
       help='Disable all "gclient runhooks" and "gclient sync" commands')
   reproduce.add_argument(
-      '-i', '--iterations', action='store', default=None, type=int,
+      '-i', '--iterations', action='store', default=10, type=int,
```
tanin47: I think we have a `while` condition where it expects iterations to be `None`. Could you remove that logic as well?

everestmz: I thought it may be useful in case we want to call reproduce to run indefinitely sometime, but that is a strange usecase anyways - I'll remove it.

tanin47: Yup, set it to 100,000 is probably considered as eternity anyway.

everestmz: Updated!

Issue #129, opened 2017-04-03
The tool crashes if testcase file has been already downloaded


Dor1s:
```bash
$ clusterfuzz reproduce -dg 6083453533814784
Reproducing testcase 6083453533814784
Downloading testcase information...

We need to authenticate you in order to get information from ClusterFuzz.

Open: <....>

Please login on the opened webpage and enter your verification code: <...>
Downloading testcase data...

Detailed logs of this run can be found in: ~/.clusterfuzz/logs.
Traceback (most recent call last):
  File "/usr/local/bin/clusterfuzz", line 11, in <module>
    sys.exit(execute())
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/main.py", line 53, in execute
    command.execute(**arg_dict)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/stackdriver_logging.py", line 123, in wrapped
    func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/commands/reproduce.py", line 250, in execute
    definition.sanitizer)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/reproducers.py", line 46, in __init__
    self.testcase_path = testcase.get_testcase_path()
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/testcase.py", line 116, in get_testcase_path
    os.makedirs(testcase_dir)
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/usr/local/google/home/mmoroz/.clusterfuzz/testcases/6083453533814784_testcase'
```

Once again:
```bash
$ clusterfuzz reproduce -dg 6083453533814784
Reproducing testcase 6083453533814784
Downloading testcase information...
Downloading testcase data...

Detailed logs of this run can be found in: ~/.clusterfuzz/logs.
Traceback (most recent call last):
  File "/usr/local/bin/clusterfuzz", line 11, in <module>
    sys.exit(execute())
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/main.py", line 53, in execute
    command.execute(**arg_dict)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/stackdriver_logging.py", line 123, in wrapped
    func(*args, **kwargs)
[...]

PR #137, opened 2017-04-03
Delete testcase file before every run

everestmz:
cc: @google/clusterfuzz 
Fixes: #129


tool/test/clusterfuzz/testcase_test.py
```diff
@@ -117,10 +117,24 @@ def test_already_downloaded(self):
       f.write('testcase exists')
     self.assertTrue(os.path.isfile(filename))
 
+    def do_wget(*unused_args):
```
tanin47: Could you merge it with the test method below.

tool/clusterfuzz/testcase.py
```diff
@@ -106,8 +107,8 @@ def get_testcase_path(self):
 
     testcase_dir = self.testcase_dir_name()
     filename = os.path.join(testcase_dir, 'testcase%s' % self.file_extension)
-    if os.path.isfile(filename):
-      return filename
+    if os.path.exists(testcase_dir):
```
tanin47: Make a common's method for delete if exists. Then, we will have only one test below.

everestmz:
Updated!

tool/test/clusterfuzz/testcase_test.py
```diff
@@ -101,27 +101,14 @@ def setUp(self):
...
-    with open(filename, 'w') as f:
-      f.write('testcase exists')
-    self.assertTrue(os.path.isfile(filename))
-
-    result = self.test.get_testcase_path()
-    self.assertEqual(result, filename)
-    self.assert_n_calls(0, [self.mock.get_stored_auth_header,
-                            self.mock.execute])
-
   def test_not_already_downloaded(self):
```
tanin47: Rename the test.

Issue #130, opened 2017-04-03
Ask users to install blackbox if it doesn't exist.


everestmz:

tanin47:
We simply raise an exception with more info.

PR #139, opened 2017-04-04
Raise an exception if blackbox is not installed

everestmz:
cc: @google/clusterfuzz 
fixes: #130


tool/clusterfuzz/reproducers.py
```diff
@@ -194,8 +194,12 @@ def __enter__(self):
         display_name = i
         break
     logger.info('Starting the blackbox window manager in a virtual display.')
-    self.blackbox = subprocess.Popen(['blackbox'],
-                                     env={'DISPLAY': display_name})
+    try:
+      self.blackbox = subprocess.Popen(['blackbox'],
+                                       env={'DISPLAY': display_name})
+    except OSError:
+      raise common.BlackboxNotInstalledError
```
tanin47: I think we should check specifically for `OSError: [Errno 2] No such file or directory` when the blackbox is not found.

everestmz: I agree that's a good idea - updated!

PR #143, opened 2017-04-05
Move job type definitions to Yaml

everestmz:
cc: @tanin47 
An initial push for a look at the structure before I write tests.


tool/clusterfuzz/supported_job_types.yml
```diff
@@ -0,0 +1,82 @@
...
+        linux_asan_d8_dbg:
+            preset: d8
+        linux_asan_d8:
+            preset: d8
+        linux_asan_d8_v8_mipsel_db:
+            preset: d8
+        linux_v8_d8_tot:
+            preset: d8
+chromium:
+        linux_asan_pdfium:
```
tanin47: Could you try if inline hash works? e.g `linux_asan_pdfium: { one: 1, two: 2 }`

It would be great if we can sort the list alphabetically.

everestmz: Yeah, that would be nice! I'll take a look

everestmz: Would we just want one long line for each entry then?

tanin47: For jobs, under chromium, the line doesn't seem to be long. Each line has only 2 or 1 entries.

Issue #144, opened 2017-04-06
SSLError: [Errno 8] _ssl.c:510: EOF occurred in violation of protocol


tanin47:
We encountered this problem before, so we switched fromr `urlfetch` to `requests` in #65.

It seems to work for us. However, one of our users has informed us that he is seeing this exception:

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/stackdriver_logging.py", line 128, in wrapped
    func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/commands/reproduce.py", line 227, in execute
    response = get_testcase_info(testcase_id)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/commands/reproduce.py", line 180, in get_testcase_info
    return json.loads(send_request(CLUSTERFUZZ_TESTCASE_INFO_URL, data).text)
  File "/usr/local/lib/python2.7/dist-packages/clusterfuzz/commands/reproduce.py", line 162, in send_request
    allow_redirects=True, data=data)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 110, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/adapters.py", line 497, in send
    raise SSLError(e, request=request)
SSLError: [Errno 8] _ssl.c:510: EOF occurred in violation of protocol
```

We will try a fix here: https://github.com/kennethreitz/requests/issues/3391

It suggests that we do:

```python
import requests
from requests.adapters import HTTPAdapter

s = requests.Session()
s.mount('https://', HTTPAdapter(max_retries=3))
```

tanin47:
I'm not sure if using HTTPAdapter will solve the problem or not. Another solution is to select the right SSL version: https://lukasa.co.uk/2013/01/Choosing_SSL_Version_In_Requests/

tanin47:
[...]

PR #145, opened 2017-04-06
Fix SSLError

tanin47:
A speculative fix for #144.


tanin47:
We've found the root cause. We don't need this fix.

PR #146, opened 2017-04-06
Another Speculative SSL fix

tanin47:
For #144


tanin47:
We've found the root cause. We don't need this fix.

PR #148, opened 2017-04-06
Add sanity checks to CI for clusterfuzz release

everestmz:
cc: @google/clusterfuzz 

This CL adds the following functionality to the CI module:
 - Clones chromium & installs depot tools
 - Installs the latest release candidate (hardcoded right now)
 - Runs all testcases inside sanity_checks.yml
 - Logs the failure or success to Stackdriver
 - Adds a new service, `auth-update`, that keeps the auth token updated so it doesn't expire in the middle of a testcase run
 - Various updates to `playbook.yml` to make everything work, like using a 100GB disk size, moving to ubuntu, ensuring that clusterfuzz user has password-less sudo, etc


.gitignore
```diff
@@ -1,5 +1,6 @@
 *.egg-info
 *.pyc
+*.py~
```
tanin47: What is this file?

ci/auth_update/main.py
```diff
@@ -0,0 +1,31 @@
+"""Keeps the authorization token in ~/.clusterfuzz/auth_header updated."""
```
tanin47: Why do we need to have a separate process for this? Specifically, why doesn't a method in main.py work?

ci/continuous_integration/ci/main.py
```diff
@@ -0,0 +1,80 @@
...
+  clone_chromium.main()
+  install_latest_release()
+  testcases = load_sanity_check_testcases()
+  for testcase in testcases:
+    if os.path.exists(CHROMIUM_OUT):
+      shutil.rmtree(CHROMIUM_OUT)
+    if os.path.exists(CLUSTERFUZZ_BUILD):
+      shutil.rmtree(CLUSTERFUZZ_BUILD)
+    stackdriver_logging.send_run(
+        testcase, 'sanity', '0.2.2rc3', run_testcase(testcase))
```
tanin47: I think we should use a decoration to capture any exception from it.

circle.yml
```diff
@@ -6,6 +6,6 @@ test:
   override:
     - python pylint.py
     - cd tool && coverage run test.py
+    - cd ci/continuous_integration && coverage run test.py
   post:
     - cd tool && coveralls
```
tanin47: This doesn't send coverage of ci/continuous_integration to coveralls. Need more research on this part.

.gitignore
everestmz: Sometimes vim doesn't clean up it's swap files properly, this just ignores them.

ci/continuous_integration/test/helpers.py
```diff
@@ -0,0 +1,85 @@
+"""Helper methods and classes to be used by all tests."""
```
tanin47: I think we should refactor this and share the code between `tool` and `ci`. For tests, we can manually modify `sys.path` in the code.

ci/continuous_integration/ci/clone_chromium.py
```diff
@@ -0,0 +1,35 @@
...
+DEPOT_TOOLS = os.path.join(HOME, 'depot_tools')
+CHROMIUM_DIR = os.path.join(HOME, 'chromium')
+CHROMIUM_SRC = os.path.join(CHROMIUM_DIR, 'src')
+
+
+def clone_chromium():
+  """Runs the correct commands to clone chromium & depot tools."""
+
+  if not os.path.exists(DEPOT_TOOLS):
+    subprocess.call([
```
tanin47: Do `shell=True` on all of them, so that they are in the same format (nicer) and is closer to what humans do.

ci/continuous_integration/ci/clone_chromium.py
```diff
@@ -0,0 +1,35 @@
...
+  if not os.path.exists(CHROMIUM_DIR):
+    os.makedirs(CHROMIUM_DIR)
+
+  subprocess.call([os.path.join(DEPOT_TOOLS, 'fetch'), '--nohooks',
+                   'chromium'], cwd=CHROMIUM_DIR)
+  subprocess.call(['build/install-build-deps.sh'], cwd=CHROMIUM_SRC)
+  subprocess.call('export PATH=$PATH:%s && gclient runhooks' % DEPOT_TOOLS,
+                  shell=True, cwd=CHROMIUM_SRC)
+
+def main():
```
tanin47: Remove this `main()`. We don't seem to need it. Move `if` into `clone_chromium`.

ci/continuous_integration/ci/clone_chromium.py
```diff
@@ -0,0 +1,35 @@
...
+  if not os.path.exists(CHROMIUM_SRC):
+    clone_chromium()
+
+if __name__ == '__main__':
+  main()
```
tanin47: Just call `clone_chromium` directly. Normally, we have `main` because its responsibility is to parse flag.

.gitignore
tanin47: Let's remove it; it's oddly specific to support only vim. (I point my swap file to `tmp`).

ci/continuous_integration/ci/main.py
everestmz: What would the decorator do? The exception would be recorded by svlogd and runit would restart the process so we don't have to worry about that.

tanin47: Ok, fair enough. This is good then.

everestmz:
Updated:
 - removed __main__ functionality from clone_chromium
 - moved helpers.py into shared folder and modified sys.path in both test.py
 - moved updating auth into the ci module
 - small changes

Still need to work out sending coverage for both test results

ci/playbook.yml
```diff
@@ -3,13 +3,26 @@
...
+    - name: Create storage disk
+      gce_pd:
+        disk_type: pd-ssd
+        zone: "us-central1-f"
+        project_id: "981641712411"
+        size_gb: 100
+        image: ubuntu-1404
+        name: "ci-test-boot"
+        state: present
+        service_account_email: "ci-test-tanin@clusterfuzz-tools.iam.gserviceaccount.com"
```
tanin47: After this CL, please switch to `ci-test@` and move the credentials file to `/auto/...`

ci/playbook.yml
```diff
@@ -3,13 +3,26 @@
...
+        credentials_file: "clusterfuzz-tools-ci-credentials.json"
     - name: Create the instance
       gce:
+        disks:
+          - name: "ci-test-boot"
+            mode: READ_WRITE
         instance_names: continuous-integration-test
-        machine_type: n1-standard-8
-        image: debian-7
+        machine_type: n1-standard-32
```
tanin47: 64 is not possible?

ci/playbook.yml
```diff
@@ -65,22 +94,23 @@
...
             dest={{item.dest}}
             force=true
             owner=root
             group=root
             mode='u=rwx,g=rwx,o=rwx'
       with_items:
-        - {'src': 'python_daemon_run', 'dest': '/etc/sv/python-daemon/run'}
-        - {'src': 'python_daemon_log_run', 'dest': '/etc/sv/python-daemon/log/run'}
-        - {'src': 'test.py', 'dest': '/python-daemon/test.py'}
+        - {'src': 'python_daemon/run', 'dest': '/etc/sv/python-daemon/run'}
```
tanin47: Do we still need this python_daemon?

ci/playbook.yml
everestmz: Not in the current region. I didn't want to switch regions along with all of this - I'll do it in another CL

tanin47: sounds good.

PR #150, opened 2017-04-07
Only allow installation with pip 9

tanin47:


tool/setup.py
```diff
@@ -14,12 +14,20 @@
 # limitations under the License.
 
 
+import pip
 from setuptools import setup
 
 
+if pip.__version__ < '8':
```
everestmz: Do we want `<= 8`, or is version 8 ok?

tanin47: 8 is ok. version 9 was only launched a few months ago.

tanin47: Second though: let's just do 9. I don't why we need to support 8.

PR #161, opened 2017-04-13
Add source directory caching

everestmz:
cc: @google/clusterfuzz


tool/clusterfuzz/common.py
```diff
@@ -34,6 +35,7 @@
 DOMAIN_NAME = 'clusterfuzz.com'
 DEBUG_PRINT = os.environ.get('CF_DEBUG')
 TERMINAL_WIDTH = get_terminal_size().columns
+SOURCE_CACHE = os.path.join(CLUSTERFUZZ_DIR, 'source_cache')
```
tanin47: Just call it cache

tool/clusterfuzz/common.py
```diff
@@ -336,3 +338,34 @@ def delete_if_exists(path):
...
+
+def get_source_directory(source_name):
+  """Returns the location of the source directory."""
+
+  source_env = '%s_SRC' % source_name.upper()
+
+  if os.environ.get(source_env):
+    return os.environ.get(source_env)
+
+  if os.path.exists(SOURCE_CACHE):
```
tanin47: This should be a separate method and more generic. We will just it for every other setting.

tanin47: I'm thinking maybe we should cache everything we ask (similar to how ssh's known host or config works)

everestmz: Do we really need a generic method here? This is the only place we really set a specific file in such a way.

everestmz: It didn't load your second comment until now. That would be interesting but the only thing we ask is the directory. Everything else is stuff we need to confirm like switching branches.

tanin47: Fair enough.

PR #163, opened 2017-04-13
[Not ready for review yet] Move to Bazel

tanin47:


tanin47:
Not feasible. It doesn't support code coverage for python. I'll summarize why.

PR #175, opened 2017-04-17
Print ENV on command-line

tanin47:


tool/clusterfuzz/commands/reproduce.py
```diff
@@ -120,10 +120,7 @@ def ensure_goma():
   if not os.path.isfile(os.path.join(goma_dir, 'goma_ctl.py')):
     raise common.GomaNotInstalledError()
 
-  common.execute(
-      'python goma_ctl.py ensure_start', goma_dir,
-      environment=os.environ.copy())
```
everestmz: Are we sure that we don't need to include the OS environment here? I think I recall it being necessary but I can't remember why...

tanin47: `python -c "import subprocess; subprocess.call('env', shell=True)"` seems to print all the envs, so `environment=os.environ.copy()` seems redundant.

PR #179, opened 2017-04-19
Fix LibfuzzerMsanBuilder

everestmz:
cc: @google/clusterfuzz


tool/clusterfuzz/binary_providers.py
```diff
@@ -395,13 +395,9 @@ class LibfuzzerMsanBuilder(ChromiumBuilder):
...
-          'gclient runhooks',
-          self.source_directory,
-          environment={
-              'GYP_DEFINES': (
-                  'clang=1 component=static_library gomadir=%s msan=1 '
-                  'msan_track_origins=2 proprietary_codecs=1 target_arch=x64 '
-                  'use_goma=1 use_prebuilt_instrumented_libraries=1'
-                  % self.goma_dir)
-          })
+      env = os.environ.copy()
```
tanin47: Actually, let's do this in `start_execute(..)`. That would be closer to how users would run if they were to run the command manually.

everestmz: Good point, it would also stop any similar issues like this

everestmz:
Updated!

PR #181, opened 2017-04-19
Add supported job types command

everestmz:
cc: @google/clusterfuzz 
Adds a command to display supported job types, so the CI knows what to filter.


tool/clusterfuzz/commands/supported_job_types.py
```diff
@@ -0,0 +1,33 @@
...
+logger = logging.getLogger('clusterfuzz')
+
+def execute():
+  """Echos all supported job types."""
+
+  logger.debug('Printing supported job types')
+
+  supported_jobs = reproduce.get_supported_jobs()
+  for category in supported_jobs:
+    logger.info('Supported %s job types:', category)
```
tanin47: Is this easily parsable by the CI? And I think we can drop the category line because it doesn't seem to provide much value.

Could you print the location of the yaml file as well? It might be easier to let CI read directly from the yaml file?

everestmz: We need that line to separate so that Ci doesn't accept a standalone-only testcase to run on chromium source. And it helps for humans. I think it's simple enough, it's just each job type on a new line.

tanin47: Actually, I take that back... if we print the output in a valid yaml format... we don't really need the yaml location.

Please use yaml's serialization for the output (instead of manually printing it ourselves).

tanin47: Sounds good. But I'd prefer it to be:

1. Construct a dict
2. Use `logging.info(yaml.dumps(the_dict))`.

It'd be more scalable if we add anything in the future.

everestmz: Updated!

tool/clusterfuzz/commands/supported_job_types.py
```diff
@@ -0,0 +1,37 @@
...
+def execute():
+  """Echos all supported job types."""
+
+  logger.debug('Printing supported job types')
+
+  supported_jobs = reproduce.get_supported_jobs()
+  to_print = {}
+  for category in supported_jobs:
+    to_print[category] = []
+    for i in supported_jobs[category]:
```
tanin47: Rename `i` to `job`

PR #192, opened 2017-04-21
Decouple supported_job_types from reproduce

everestmz:
cc: @google/clusterfuzz 

Read directly from the yaml rather than going through lots of needless parsing in `reproduce.get_supported_jobs()`.


tool/clusterfuzz/commands/supported_job_types.py
```diff
@@ -27,7 +27,9 @@ def execute():
 
   logger.debug('Printing supported job types')
 
-  supported_jobs = reproduce.get_supported_jobs()
+  with open(common.get_resource(
```
tanin47: It's probably little things... but I'd prefer this in `common.py`.

PR #211, opened 2017-04-26
Support linux_ubsan_vptr_content_shell_drt

tanin47:


tool/clusterfuzz/reproducers.py
```diff
@@ -274,7 +274,7 @@ def __enter__(self):
         raise common.BlackboxNotInstalledError
       raise
 
-    time.sleep(30)
+    time.sleep(3)
```
tanin47: It seems 1 second is enough. So, setting it to 3 just to be sure.

PR #224, opened 2017-05-02
Remove disable-gclient; Rename blackbox

tanin47:
- Rename Blackbox to Xvfb. Blackbox is just a window manager. We were confused between the two terms.
- `--disable-gclient` doesn't make sense anymore. User can just use `--current` which means user ensure that the repo is in a good state for building.


tool/clusterfuzz/main.py
```diff
@@ -51,14 +53,11 @@ def execute(argv=None):
...
-      '--disable-gclient-commands', action='store_true', default=False,
-      help='Disable all "gclient runhooks" and "gclient sync" commands')
-  reproduce.add_argument(
       '-i', '--iterations', action='store', default=10, type=int,
       help='Specify the number of times to attempt reproduction')
   reproduce.add_argument(
-      '-db', '--disable-blackbox', action='store_true', default=False,
-      help='Disable running testcases in a virtual display with Blackbox')
+      '-db', '--disable-xvfb', action='store_true', default=False,
+      help='Disable running testcases in a frame buffer')
```
mbarbella: Nit: add virtual

PR #226, opened 2017-05-02
Expand path when user use '.'

tanin47:
- When asking user for CHROMIUM_SRC, we allow user to use `.`. We've found that users usually run within the chrome directory.
- Remove source cache; it seems redundant with using the environment variable. And it simplifies code.


tool/clusterfuzz/common.py
```diff
@@ -386,24 +390,11 @@ def get_source_directory(source_name):
...
              {'name': source_name, 'env_name': source_env})
-  source_directory = os.path.expanduser(
-      ask(message, 'Please enter a valid directory',
-          lambda x: x and os.path.isdir(os.path.expanduser(x))))
 
-  with open(SOURCE_CACHE, 'w') as f:
-    source_locations[source_env] = source_directory
-    f.write(yaml.dump(source_locations))
+  source_directory = os.path.abspath(os.path.expanduser(
+      ask(message, 'Please enter a valid directory', is_valid_dir)))
```
mbarbella: It might be slightly cleaner to replace is_valid_dir with something like get_expanded_directory, and have that return None in the event that it doesn't exist. This isn't too complicated as-is, though, so fine to leave it.

tanin47: Yeah, I think there's a redundancy here because `ask` only takes `validate_fn` which validates the input. It isn't responsible for transforming the input.

What you describe is validation + transformation. Maybe we should do that instead. Let me see how much change is involved.

Thanks for the suggestion!

tanin47: Talked offline. Marty suggested using the function twice instead. That makes sense. Will make the change.

PR #232, opened 2017-05-04
Add linux job types.

inferno-chromium:


tool/clusterfuzz/binary_providers.py
```diff
@@ -334,9 +334,7 @@ def __init__(self, testcase, binary_definition, current, goma_dir,
     self.name = 'V8'
 
   def pre_build_steps(self):
-    if not self.current:
-      common.execute('gclient', 'runhooks', self.source_directory)
-    common.execute('gypfiles/gyp_v8', '', self.source_directory)
```
tanin47: I didn't know we didn't need this. Thank you.

tool/clusterfuzz/binary_providers.py
```diff
@@ -359,8 +357,8 @@ def __init__(self, testcase, binary_definition, current, goma_dir,
     self.name = 'chromium'
 
   def pre_build_steps(self):
+    common.execute('gclient', 'runhooks', self.source_directory)
```
tanin47: What is the reason for moving this out?

`--current` means that the user ensures the repo is in a good state and ready to be built.

inferno-chromium: If the repo is not in good state, it fails with a bad silent error. Running runhooks just brings the dependencies, there is nothing custom in there, like custom clang.

tanin47: Ok, that's fair. The deal breaker is `gclient sync` anyway, not `runhooks`.

PR #233, opened 2017-05-04
Modify gn args and args with editor.

tanin47:
Moving certain things around ... so the change is a bit big.


cmd-editor/cmd_editor/editor.py
```diff
@@ -21,13 +21,29 @@
...
   except subprocess.CalledProcessError:
     raise Exception('%s is not found. Please ensure it is in PATH.' % binary)
 
 
+def add_comment(content, comment):
+  """Add comment to the content."""
+  if not comment:
+    return content
+
+  comment_block = '\n'.join(['# ' + line for line in comment.split('\n')])
```
inferno-chromium: nit: '# %s' * line

PR #236, opened 2017-05-05
Fixes

inferno-chromium:
- Make sure user is on a branch.
- Clean up temporary build directory.
- Add clang update step for V8 builds as well.
- Fix d8 job names.
- Make TSan jobs work with longer timeout.
- Make LSan jobs work with need of some sleep after shutdown.
- Add Chomium d8 job types.
- Fix bug with symbolizer hanging due to null input.
- Fix log file path.


tool/clusterfuzz/binary_providers.py
```diff
@@ -300,14 +314,21 @@ def get_build_directory(self):
...
       self.source_directory = common.get_source_directory(self.name)
+
     if not self.current:
       self.checkout_source_by_sha()
+
     self.build_directory = self.out_dir_name()
     self.build_target()
 
+    # Cleanup any artifacts in temporary build directory.
+    common.delete_if_exists(CLUSTERFUZZ_BUILDS_DIR)
```
tanin47: This is `~/.clusterfuzz/builds` which contains downloaded builds. Why do we remove it?

tool/clusterfuzz/binary_providers.py
```diff
@@ -279,8 +289,12 @@ def get_goma_cores(self):
 
   def build_target(self):
     """Build the correct revision in the source directory."""
+    if self.get_current_branch() == 'HEAD':
```
tanin47: If I understand it correctly, this forces user to create a branch. What is the rational behind this?

tool/clusterfuzz/binary_providers.py
inferno-chromium: We are not reusing any builds, we always gsutil cp on it. It fills up disk space pretty quick, we should cleanup this secret location. A user knows about out/ directories, but not this one.

tanin47: Discussed offline. We have been thinking about this but haven't implemented it yet. I think the more effective way is to rotate the build folders correctly. Otherwise, we would download a build everytime we run in the case of missing args.gn or run with `-b download`.

There should be a python package that achieves what we want here.

tool/clusterfuzz/binary_providers.py
tanin47: Discussed offline. This is good to merge.

tool/clusterfuzz/binary_providers.py
tanin47: Could you remove this part in the mean time (or move it the top of the method)? Otherwise, this wouldn't work with the "args.gn in build archive" case.Thank you!

PR #237, opened 2017-05-05
Remove disable-gl-drawing-for-tests flag.

inferno-chromium:


tanin47:
Talked offline. I misunderstood it. We don't really need to disable GL locally even when running inside xvfb. Let's merge. Thank you!

PR #240, opened 2017-05-06
Make testcase dependent on layouttest work (e.g. 4635463068680192).

inferno-chromium:


inferno-chromium:
Yes --log-net-log seems to fail often, i have seen this in a ton of testcases now. Will discuss more next week.

PR #245, opened 2017-05-08
Fixes

inferno-chromium:
- Add linux_msan_d8
- Add _standalone prefix for d8.
- Rename CfiChromium to CFIChromium
- Actually use MSANBuilderChromium


tool/clusterfuzz/binary_providers.py
```diff
@@ -376,12 +376,12 @@ def pre_build_steps(self):
                      self.source_directory)
 
 
-class CfiChromiumBuilder(ChromiumBuilder):
+class CFIChromiumBuilder(ChromiumBuilder):
```
tanin47: Could you use `Cfi` instead of `CFI`? It becomes problematic when 2 abbreviation are next to each other, e.g. `MSANCFI`.

PR #249, opened 2017-05-09
Prompt for installing 32-bit libs.

inferno-chromium:


tool/clusterfuzz/common.py
```diff
@@ -326,6 +326,19 @@ def execute(binary, args, cwd, print_command=True, print_output=True,
...
+def execute_with_shell(binary, args, cwd):
+  """Execute command with os.system."""
+  check_binary(binary, cwd)
+
+  command = os.path.join(cwd, binary)
+  if args:
+    command += ' ' + args
+
+  log = ('Running %s', command)
+  logger.info(*log)
```
tanin47: Can we remove `log`?

inferno-chromium: No it is needed, otherwise we wouldn't know what command is being run.

tanin47: Oh, I mean we can do `logger.info('Running %s', command)`?

PR #250, opened 2017-05-09
Prompt for installing 32-bit libs.

inferno-chromium:


inferno-chromium:
Redo, https://github.com/google/clusterfuzz-tools/pull/249, which looks stuck on coverage.

PR #260, opened 2017-05-10
Sleep between every testcase.

tanin47:
- Sleep between each testcase. We encounter the git's rate-limit, and it goes through all testcases extremely quickly.
- Ignore `missing-docstring` on `setUp`


ci/continuous_integration/daemon/main.py
```diff
@@ -181,11 +182,13 @@ def main():
 
   for testcase in load_sanity_check_testcases():
     reset_and_run_testcase(testcase, 'sanity', release)
+    time.sleep(30)
```
inferno-chromium: Move 30 to a global var SLEEP_TIMEOUT

tanin47: Done. Thanks.

PR #264, opened 2017-05-18
Use python dbg, so we can debug when the CI hangs.

tanin47:
I've found that the CI hangs since May 12. I don't know why. This allows me to debug when it hangs.


ci/playbook.yml
```diff
@@ -67,6 +67,8 @@
     - apt: name={{item}} state=latest
       with_items:
         - python-dev
```
inferno-chromium: can we sort this list :)

tanin47: I'll do it in the next CL. Thank you.

PR #265, opened 2017-05-18
Ignore webbrowser's error; ignore error when stopping the service

tanin47:


ci/playbook.yml
```diff
@@ -66,18 +66,18 @@
         - soft
     - apt: name={{item}} state=latest
       with_items:
-        - python-dev
-        - python2.7-dbg
+        - blackbox
```
tanin47: Mere sorting. I didn't add anything here.

ci/playbook.yml
```diff
@@ -93,7 +93,8 @@
   connection: ssh
   become: true
   tasks:
-    - shell: "sv stop python-daemon"
+    - command: "sv stop python-daemon"
+      ignore_errors: yes
```
tanin47: We don't care about the error. Sometimes the service doesn't exist.

tool/clusterfuzz/commands/reproduce.py
```diff
@@ -59,6 +59,7 @@ def __enter__(self):
   def __exit__(self, unused_type, unused_value, unused_traceback):
     os.dup2(self.stdout, 1)
     os.dup2(self.stderr, 2)
+    return True
```
tanin47: Returning true ignores the error.

PR #279, opened 2017-05-22
Print stderr on screen

tanin47:
* Refactor the way we transform the output before printing to screen.
* Print stderr on screen.
* Highlight the running command in green. It looks nicer.


tool/clusterfuzz/output_transformer.py
```diff
@@ -0,0 +1,82 @@
...
+    """Replace the previous line and print output."""
+    if '\n' in string:
+      tokens = string.split('\n')
+      self.current_line += tokens[0]
+      current_line_size = len(self.current_line)
+
+      if current_line_size < self.previous_line_size:
+        self.current_line += ' ' * (self.previous_line_size - current_line_size)
+
+      output_fn('\b' * self.previous_line_size)
```
tanin47: This is more resilient than `\r`. `\r` fails if the screen width is too narrow. This way doesn't fail.

tool/tests/clusterfuzz/output_transformer_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+    transformer = output_transformer.Ninja()
+    transformer.process('aaaaa\n', output_fn)
+    transformer.process('bbb', output_fn)
+    transformer.process('\nccc', output_fn)
+    transformer.process('c\ndddddd', output_fn)
+    transformer.flush(output_fn)
+
+    self.assertEqual(
+        'aaaaa\b\b\b\b\bbbb  \b\b\b\b\bcccc \b\b\b\b\bdddddd\n', self.output)
```
tanin47: `\b` is a backspace character.

Surprisingly, `\b` doesn't erase a character. It merely moves the cursor back. Therefore, we need to add trailing space if the line is too short (to erase the character from the previous line).

tool/clusterfuzz/common.py
```diff
@@ -27,14 +28,19 @@
 
 from backports.shutil_get_terminal_size import get_terminal_size
 from clusterfuzz import local_logging
+from clusterfuzz import output_transformer
+
+
+BASH_GREEN_MARKER = '\033[92m'
```
inferno-chromium: Can you add os.name == 'posix' check for both of these. best of keep code cross-platform as we go :)

tanin47: Done.

PR #280, opened 2017-05-23
Add disable-gclient back

tanin47:
We need to bring this back because `--current` seems independent from `--disable-gclient`.


tool/tests/clusterfuzz/binary_providers_test.py
```diff
@@ -174,7 +174,7 @@ def test_parameter_not_set_valid_source(self):
                          gn_args=None)
     binary_definition = mock.Mock(source_var='V8_SRC')
     provider = binary_providers.V8Builder(
-        testcase, binary_definition, False, '/goma/dir', None, False)
+        testcase, binary_definition, False, '/goma/dir', None, False, False)
```
inferno-chromium: nit: these tests should pass the argument name explicitly just like reproduce_test.py.

tanin47: I was thinking about this as well. But let's do this separately. It's going to be a lot of changes.

PR #282, opened 2017-05-23
Add retrying; change color.

tanin47:
* Add retrying to http request.
* Change the color from green to blue.
* Add the color to an expected exception.


inferno-chromium:
Thanks a lot!

PR #292, opened 2017-05-30
Exclude --user-data-dir from all jobs except chrome and content shell.

tanin47:


tool/clusterfuzz/resources/supported_job_types.yml
```diff
@@ -31,15 +31,18 @@ presets: #Any parameters in a preset can be overridden in the actual definition
             reproducer: LinuxChromeJob
             binary: chrome
             target: chromium_builder_asan
+            should_have_user_data_dir: true
```
tanin47: Here's how we specify if we will force the argument `--user-data-dir`.

tool/clusterfuzz/reproducers.py
```diff
@@ -119,6 +121,33 @@ def serialize_sanitizer_options(options):
   return ':'.join(pairs)
 
 
+def ensure_user_data_dir_if_needed(args, should_have_user_data_dir):
+  """Ensure the right user-data-dir."""
+  if not should_have_user_data_dir and USER_DATA_DIR_ARG not in args:
```
tanin47: Here's the most important part.

If `--user-data-dir` exists in `minimized_args` OR `should_have_user_data_dir` is true, we will force `--user-data-dir`.

tool/tests/clusterfuzz/reproducers_test.py
```diff
@@ -762,3 +768,63 @@ def test_similar(self):
...
+  def test_doing_nothing(self):
+    """Test doing nothing."""
+    self.assertEqual(
+        'something',
+        reproducers.ensure_user_data_dir_if_needed('something', False))
+
+  def test_add_because_it_should(self):
+    """Test adding arg because it should have."""
+    self.assertEqual(
+        'something --user-data-dir=%s' % reproducers.USER_DATA_DIR_PATH,
```
inferno-chromium: can we add a test with some initial arguments other than --user-data-dir as well.

tool/clusterfuzz/reproducers.py
```diff
@@ -119,6 +121,33 @@ def serialize_sanitizer_options(options):
   return ':'.join(pairs)
 
 
+def ensure_user_data_dir_if_needed(args, should_have_user_data_dir):
```
tanin47: It's not exactly adding. In certain case, we might correct the path because `--user-data-dir` already existed.

tool/clusterfuzz/resources/supported_job_types.yml
```diff
@@ -31,15 +31,18 @@ presets: #Any parameters in a preset can be overridden in the actual definition
...
         linux_content_shell:
             preset: linux_chromium
             binary: content_shell
         linux_filter_stub:
             preset: linux_chromium
             binary: filter_fuzz_stub
+            should_have_user_data_dir: false
```
tanin47: Done. Thank you for the suggestion.

tool/tests/clusterfuzz/reproducers_test.py
tanin47: `something` is the initial argument because `def ensure_user_data_dir_if_needed(args, require_user_data_dir):` takes the full arguments.

tool/clusterfuzz/commands/reproduce.py
```diff
@@ -159,9 +159,13 @@ def build_binary_definition(job_definition, presets):
...
-      builders[result['builder']], result['source'],
-      reproducer_map[result['reproducer']], result.get('binary'),
-      result.get('sanitizer'), result.get('target'))
+      builder=builders[result['builder']],
+      source_var=result['source'],
+      reproducer=reproducer_map[result['reproducer']],
+      binary_name=result.get('binary'),
+      sanitizer=result.get('sanitizer'),
+      target=result.get('target'),
+      should_have_user_data_dir=result.get('should_have_user_data_dir', False))
```
tanin47: I've changed it to `require_user_data_dir`.

PR #297, opened 2017-05-31
Avoid printing stacktrace twice; hide wget's warnings

tanin47:
- Avoid printing stacktrace twice in some places.
- Raise `CommandFailedError` instead of `sys.exit` because we don't really need to see the stacktrace when a command fails
- Use `curl` instead of `wget`. Fun fact: with `wget`, we can only either print all messages or print nothing at all; this is ridiculous (https://serverfault.com/questions/70889/wget-to-print-errors-but-nothing-otherwise).


tool/clusterfuzz/testcase.py
```diff
@@ -122,19 +122,17 @@ def get_testcase_path(self):
     testcase_dir = self.testcase_dir_name()
     filename = os.path.join(testcase_dir, 'testcase%s' % self.file_extension)
     common.delete_if_exists(testcase_dir)
+    os.makedirs(testcase_dir)
 
     logger.info('Downloading testcase data...')
 
-    if not os.path.exists(common.CLUSTERFUZZ_TESTCASES_DIR):
```
tanin47: `testcase_dir` includes `common.CLUSTERFUZZ_TESTCASES_DIR`, so this is unnecessary.

tool/clusterfuzz/reproducers.py
```diff
@@ -276,9 +276,6 @@ def reproduce(self, iteration_max):
     while iterations <= iteration_max:
       _, output = self.reproduce_crash()
 
-      print
```
tanin47: The stacktrace is already printed. We don't need to do `logger.info` again.

tool/clusterfuzz/reproducers.py
```diff
@@ -470,6 +467,7 @@ def reproduce_crash(self):
       if self.gestures:
         self.run_gestures(process, display_name)
 
-      err, out = common.wait_execute(process, exit_on_error=False,
-                                     timeout=TEST_TIMEOUT)
+      err, out = common.wait_execute(
+          process, exit_on_error=False, timeout=TEST_TIMEOUT,
+          print_output=False)
```
tanin47: For this case, the stacktrace is printed after being symbolized... so we don't need to print it here.

tool/clusterfuzz/common.py
```diff
@@ -233,6 +231,14 @@ def __init__(self):
         'commit or stash these changes and re-run this tool.')
 
 
+class CommandFailedError(ExpectedException):
+  """An exception raised when the command doesn't return 0."""
+
+  def __init__(self, command, returncode):
+    super(CommandFailedError, self).__init__(
+        '`%s` failed with the return code is %s' % (command, returncode))
```
mbarbella: Nit: -is

tanin47: I can't spell. #covfefe

inferno-chromium:
Thanks!

Issue #278, opened 2017-05-22
Solve the problem with subprocess hanging


tanin47:
http://tanin.nanakorn.com/blogs/373

tanin47:
Here's one workaround:

```
import subprocess
import tempfile
import time

with tempfile.NamedTemporaryFile() as stderr:
  proc = subprocess.Popen(
      'python yes.py', shell=True, stdout=subprocess.PIPE, stderr=stderr)

  count = 0
  for chunk in iter(lambda: proc.stdout.read(100), b''):
    count += len(chunk)

  thefile = open(stderr.name)
  thefile.seek(0)

  while True:
    chunk = thefile.read(100)
    if not chunk:
      if proc.poll() is None:
        time.sleep(0.1)
        print 'sleep'
      else:
        break
    else:
      count += len(chunk)
      print count

  thefile.close()

print 'Done'
```

tanin47:
This might not worth fixing. It only impacts the CI and is rare.

tanin47:
Here's where `python-daemon` is stuck:

```
#3 Frame 0x7f7ce8eae420, for file /usr/lib/python2.7/subprocess.py, line 476, in _eintr_retry_call (func=<built-in function waitpid>, args=(63131, 0))
    return func(*args)
#7 Frame 0x7f7ce8eae608, for file /usr/lib/python2.7/subprocess.py, line 1376, in wait (self=<Popen(_child_created=True, returncode=None, stdout=None, stdin=None, pid=63131, stderr=None, universal_newlines=False) at remote 0x7f7ce8e6bbd0>)
    pid, sts = _eintr_retry_call(os.waitpid, self.pid, 0)
#10 Frame 0x1ace500, for file /usr/lib/python2.7/subprocess.py, line 522, in call (popenargs=('/python-daemon/clusterfuzz reproduce 5166[redacted]9248',), kwargs={'shell': True, 'cwd': '/home/clusterfuzz', 'env': {'GOMA_GCE_SERVICE_ACCOUNT': 'default', 'CF_QUIET': '1', 'SHLVL': '0', 'PWD': '/etc/sv/python-daemon', 'USER': 'CI', 'PATH': '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/home/clusterfuzz/depot_tools', 'CHROMIUM_SRC': '/home/clusterfuzz/chromium/src'}})
[...]

PR #298, opened 2017-05-31
Speculative fix for hang on CI

tanin47:
Avoid doing `stderr.read` and use `communicate()` to get stderr instead. I hope this will solve #278.


tanin47:
I'm going to merge this because I want to see if the hang problem is solved by tomorrow.

PR #315, opened 2017-06-05
Print the stacktrace after symbolizing as well

tanin47:
We print unsymbolized stacktrace and symbolized stacktrace on screen.


inferno-chromium:
Maybe markers to seperate symbolized and unsymbolized stacks. or only symbolized ?

tanin47:
They are separated by the blue `Running: ... asan_symbolizer.py...`; it looks good enough. But we can  revisit.

inferno-chromium:
Sounds good.

Issue #299, opened 2017-05-31
There seems to be a stale `clusterfuzz` process on CI


tanin47:
This can come from the fact that we redeploy the CI; the daemon is killed... but the chrome process is not killed.

inferno-chromium:
i think we should do binary process cleanup after running every testcase (just like CF does, but maybe lighter version of that). that way, we dont leak bad state from one testcase into another (from stuck stale process).

tanin47:
We are using process group and kill every process in the group. We'll see if the problem with the child process occurs. If it occurs, we'll implement the cleanup.

tanin47:
Ok, this is because, when we kill python-daemon with `sv stop python-daemon`, which sends `SIGTERM` to python-daemon, its children are not killed. Therefore, we'll see `python2.7 /python-daemon/clusterfuzz reproduce 4782259122208768` being alive. And, somehow, `atexit.register` doesn't work.

This might cause a problem because 2 of `clusterfuzz` processes might be running, and it might try to checkout or run `gclient sync` at the same time.

Chrome or content_shell processes don't have any issue.

tanin47:
The problem should go away with #317.

PR #317, opened 2017-06-05
Kill a child process when the daemon is killed

tanin47:
- Make `/python-daemon-data` for the file generated by the daemon (e.g. `last_pid`)
- When running a new process, we read `last_pid`, kill the process, store the new process in `last_pid`. This guarantees that 2 of `clusterfuzz` processes never run simultaneously.

This solves #299.


ci/continuous_integration/daemon/main.py
```diff
@@ -75,8 +90,43 @@ def call(cmd, cwd='.', env=None, capture=False):
...
+    out, _ = proc.communicate()
+
+    if proc.returncode != 0:
+      raise subprocess.CalledProcessError(
+          returncode=proc.returncode, cmd=cmd, output=out)
+
+    return out
+
+
+def store_last_pid(pid):
```
inferno-chromium: Should store_last_pid and kill_last_pid become member of Popen class.

tanin47: I prefer a module method in general because it's easier to mock. In this specific case, the methods don't need `self`. But I think it might make sense to move all of these to a separate module.

tanin47: Done. Thank you!

Issue #318, opened 2017-06-05
User answering 'no' to the checkout question because user wants to use the current state of the git repo.


tanin47:
We can use the current state of the repo if the user says no.

Currently, we raise an exception.

tanin47:
After thinking about it, the trade-off isn't good enough.

We should always respect the options and shouldn't allow user to change the option value mid-way. User can easily re-run the command with the new set of options.

I'll improve the error message on this.

PR #321, opened 2017-06-05
Handle minimization failure

tanin47:
When minimization fails, there will be no `Running command: ` in the stacktrace. We'll raise an elaborated exception explaining why and what user should do.

Previously, we would raise an exception saying there's no `Running command: `.

This improves #318,


mbarbella:
I don't think this is the right way to handle this case, but I might be missing something. Ideally it shouldn't matter if minimization has finished or not when we try to reproduce, especially in the event that a bug in CF prevents it from finishing.

tanin47:
Here's an example of the failed testcase:

```
CI unsuccessfully finished running reproduce with testcase=5432[redacted]5920, build_type=chromium, current=False, and goma=enabled Traceback (most recent call last): 
...
Exception: The stacktrace doesn't contain a line starting with 'Running command: '
```

The testcase is too new. Therefore, `Running command: ` is not there, and there's no command-line with the testcase. It seems to occur quite often with the libfuzzer testcases.

I'll create an issue internally to discuss about this.

mbarbella:
Sure, probably worth discussing it more. I do remember that we had these issues when "Running command:" was missing. I'd just rather handle it in a more robust way if we can think of one.

tanin47:
In any case, if `Running command: ` is missing, we still need to inform user why it happens anyway. Its explanation is better than the current explanation.

I wonder if we should merge this now anyway because we will solve the `Running command: ` separately. And I agree that clusterfuzz could provide a more robust way for the reproducing command.

mbarbella:
That's reasonable. It's better than the current behavior. I'll take a closer look, but I still think we should come up with a longer-term solution.

tanin47:
I agree. Let's discuss more in the internal issue. Thanks.

PR #324, opened 2017-06-06
Enable full debug symbols

tanin47:
- Build with debug symbols. 
- Refactor some tests.

Running with gdb will come later.


inferno-chromium:
@oliverchang, can you review this.

tool/clusterfuzz/binary_providers.py
```diff
@@ -98,6 +98,26 @@ def get_current_sha(source_dir):
   return current_sha.strip()
 
 
+def setup_debug_symbol_if_needed(gn_args, enable_debug):
+  """Setup debug symbol if enable_debug is true. See: crbug.com/692620"""
+  if not enable_debug:
+    return gn_args
+
+  gn_args['sanitizer_keep_symbols'] = 'true'
+  gn_args['is_debug'] = 'true'
```
tanin47: Here's the important part. We set the 2 params in args.gn.

tool/tests/clusterfuzz/main_test.py
```diff
@@ -31,47 +31,20 @@ def setUp(self):
...
     """Test parse reproduce command."""
-    main.execute(['reproduce', '1234', '--disable-xvfb'])
-    main.execute(['reproduce', '1234', '-j', '25', '--current'])
-    main.execute(['reproduce', '1234', '--disable-goma', '--build', 'download'])
-    main.execute(['reproduce', '1234', '--current', '--build', 'standalone'])
-    main.execute(['reproduce', '1234', '--build', 'chromium', '-i', '500'])
-    main.execute(['reproduce', '1234', '--target-args', '--test --test2'])
-    main.execute(['reproduce', '1234', '--edit-mode'])
-    main.execute(['reproduce', '1234', '--disable-gclient'])
+    main.execute(['reproduce', '1234'])
```
tanin47: We shorten the test because we don't really need to test every flag independently.

tool/clusterfuzz/binary_providers.py
oliverchang: not sure if it still matters, but sanitizer_keep_symbols was added fairly recently (https://chromium.googlesource.com/chromium/src/+/2a7ed44b0ffb4f4e8f70dd5c0c82ad0046d84964). for older testcases (> 1 month), attempting to do so might cause GN to complain.

tanin47:
That's probably ok since there's no other way to enable this. We'll just let it fail.

tanin47:
Thank you! Will merge.

Issue #326, opened 2017-06-07
When the testcase file is 0 bytes, we can't download the file.


tanin47:
It seems curl doesn't support downloading an empty file. https://stackoverflow.com/questions/30537292/curl-wont-download-empty-file, https://superuser.com/questions/43278/force-curl-to-download-an-empty-file

The bug is fixed in 7.42.0 (https://curl.haxx.se/changes.html#7_42_0). But, sadly, Goobuntu still uses curl 7.35.0.

tanin47:
We'll switch back to wget and tolerate the fact that wget prints something on stderr even when there's no error. Because this requires less code.

tanin47:
Solved by #327

PR #327, opened 2017-06-07
Fix curl not downloading an empty file

tanin47:
- We fix it by switching back to wget. We'll tolerate the fact that wget prints something on stderr even when there's no error. The added code to handle empty file with curl isn't worth it.
- Add caching on CircleCI to make it run even faster.

This solves #326.


tool/clusterfuzz/testcase.py
```diff
@@ -127,12 +128,13 @@ def get_testcase_path(self):
     logger.info('Downloading testcase data...')
 
     auth_header = common.get_stored_auth_header()
+    # Do not use curl because curl doesn't support downloading an empty file.
+    # See: https://github.com/google/clusterfuzz-tools/issues/326
     args = (
-        '--remote-name --remote-header-name --retry-max-time 30 --retry 5 '
-        '--silent --show-error --fail --location '
-        '-H "Authorization: %s" %s' %
+        '--no-verbose --waitretry=80 --retry-connrefused --content-disposition '
```
inferno-chromium: Why 80, can we round to 100 and move to global variable.

tanin47: No particular reason. It's 80 seconds. Will change to 100 and move to global.

PR #328, opened 2017-06-07
--enable-debug support in reproducers.

tanin47:
- Jump into `gdb` when `--enable-debug` is used.
- `start_execute` to support different modes of `stdin`. Specifically, all `xdotool` commands' stdins need to be blocked. Otherwise, it wouldn't work with gdb.


tool/clusterfuzz/common.py
```diff
@@ -386,30 +386,72 @@ def check_binary(binary, cwd):
...
+  def get(self):
+    """Get the stdin handler for Popen."""
+    raise NotImplementedError
+
+  def update_cmd_log(self, cmd):
+    """Modify the command to represent the stdin in logs."""
+    raise NotImplementedError
+
+
+class BlockStdin(Stdin):
```
tanin47: This is needed when running `gdb` with gestures. Otherwise, what user types would go into `xdotool` instead.

tool/clusterfuzz/common.py
```diff
@@ -386,30 +386,72 @@ def check_binary(binary, cwd):
...
+    return subprocess.PIPE
+
+  def update_cmd_log(self, cmd):
+    """Return cmd because blocking input doesn't alter
+      the command."""
+    return cmd
+
+
+class UserStdin(Stdin):
+  """Accept user's input as stdin."""
```
tanin47: This is needed for `gdb` and is the default.

tool/clusterfuzz/common.py
```diff
@@ -386,30 +386,72 @@ def check_binary(binary, cwd):
...
+    return None
+
+  def update_cmd_log(self, cmd):
+    """Return cmd because accepting user's input doesn't alter
+      the command."""
+    return cmd
+
+
+class StringStdin(Stdin):
+  """Send a string as stdin."""
```
tanin47: This is needed for `asan_symbolize.py`.

tool/clusterfuzz/common.py
```diff
@@ -386,30 +386,72 @@ def check_binary(binary, cwd):
...
+    self.stdin.seek(0)
+
+  def get(self):
+    """Get the file handler for the string."""
+    return self.stdin
 
-  return stdin, stdin_log
+  def update_cmd_log(self, cmd):
+    """Add the input filename to the command."""
+    return '%s < %s' % (cmd, self.stdin.name)
```
tanin47: We make this method... so that the printed command can be copied and pasted.

tool/clusterfuzz/binary_providers.py
```diff
@@ -100,13 +100,17 @@ def get_current_sha(source_dir):
...
+def setup_debug_symbol_if_needed(gn_args, sanitizer, enable_debug):
   """Setup debug symbol if enable_debug is true. See: crbug.com/692620"""
   if not enable_debug:
     return gn_args
 
   gn_args['sanitizer_keep_symbols'] = 'true'
-  gn_args['is_debug'] = 'true'
+  if sanitizer == 'MSAN':
+    # MSAN raises an exception when is_debug = true.
+    gn_args['symbol_level'] = '2'
```
inferno-chromium: Shouldn't gn_args['symbol_level'] = '2' be done for all case, why only MSan.

inferno-chromium:
+cc Oliver as well for review.

tool/clusterfuzz/binary_providers.py
tanin47: `is_debug` seems to work for ASAN and UBSAN. 

I actually don't know what the difference is between `symbol_level = 2` and `is_debug =true`. @oliverchang ?

Issue #330, opened 2017-06-09
Error when the job type is not supported because the stacktrace cannot be parsed.


tanin47:
```
[redacted] unsuccessfully finished running reproduce with testcase=[redacted], build_type=chromium, current=False, and goma=enabled
Traceback (most recent call last):
  File "/usr/local/google/home/[redacted]/.pex/code/[redacted]/clusterfuzz/stackdriver_logging.py", line 120, in wrapped
    func(**params)
  File "/usr/local/google/home/[redacted]/.pex/code/[redacted]/clusterfuzz/commands/reproduce.py", line 240, in execute
    current_testcase = testcase.Testcase(response)
  File "/usr/local/google/home/[redacted]/.pex/code/[redacted]/clusterfuzz/testcase.py", line 79, in __init__
    self.environment, self.reproduction_args = self.get_environment_and_args()
  File "/usr/local/google/home/[redacted]/.pex/code/[redacted]/clusterfuzz/testcase.py", line 58, in get_environment_and_args
    name, value = line.split(' = ')
ValueError: need more than 1 value to unpack
```

The job type `android_asan_chrome_latest`'s stacktrace contains this line `	[Environment] Patch level: 2017-05-01`.

PR #331, opened 2017-06-09
Fix invalid [Environment] in Stacktrace and invalid stderr.

tanin47:
See #330 and #332.


tool/clusterfuzz/testcase.py
```diff
@@ -57,7 +57,11 @@ def get_environment_and_args(self):
     for line in stacktrace_lines:
       if '[Environment] ' in line:
         line = line.replace('[Environment] ', '')
-        name, value = line.split(' = ')
+        tokens = line.split(' = ')
```
inferno-chromium: Can you add split(' = ', 1) to make sure we have only one split.

tanin47: Fixed. Thanks for pointing it out.

PR #355, opened 2017-06-14
Fix a bug on CI

tanin47:


ci/continuous_integration/daemon/main.py
```diff
@@ -80,7 +80,7 @@ def run_testcase(testcase_id):
             'GOMA_GCE_SERVICE_ACCOUNT': 'default',
             'PATH': '%s:%s' % (os.environ['PATH'], DEPOT_TOOLS)
         }
-    )
+    )[0]
```
tanin47: This is where the bug is. I thought it returned a return code.

inferno-chromium: We use output from both stdout, stderr, so both should be used.

tanin47: On the CI, we don't care about stderr (yet). When running the reproduce tool, we only care about the return code.

Issue #359, opened 2017-06-14
os.rename doesn't work when moving files across file systems.


tanin47:
Here's the stacktrace:

```
"/usr/local/google/home/[redacted]/.pex/code/f77094091354d71b982764e6849685d731467916/clusterfuzz/stackdriver_logging.py", line 124, in wrapped func(**params) File "/usr/local/google/home/[redacted]/.pex/code/f77094091354d71b982764e6849685d731467916/clusterfuzz/commands/reproduce.py", line 276, in execute reproducer.reproduce(iterations) File "/usr/local/google/home/[redacted]/.pex/code/f77094091354d71b982764e6849685d731467916/clusterfuzz/reproducers.py", line 337, in reproduce self.pre_build_steps() File "/usr/local/google/home/[redacted]/.pex/code/f77094091354d71b982764e6849685d731467916/clusterfuzz/reproducers.py", line 473, in pre_build_steps self.testcase_path, self.original_testcase_path, self.source_directory) File "/usr/local/google/home/[redacted]/.pex/code/f77094091354d71b982764e6849685d731467916/clusterfuzz/reproducers.py", line 149, in update_testcase_path_in_layout_test os.rename(testcase_path, new_testcase_path) OSError: [Errno 18] Invalid cross-device link
```

@wangxianzhu suggests that we should use `shutils.move()` instead. He also has the patch ready. I'll wait for him to create a PR.

PR #360, opened 2017-06-14
Use shutil.move() instead of os.rename() to move test file

wangxianzhu:
os.rename() requires the source and dest on the same file system.
shutil.move() can handle files on different file systems.

For issue #359.


tanin47:
Somehow, coveralls doesn't work with PR from a fork. I'll figure it out later.

PR #361, opened 2017-06-14
Test PR from a fork.

tanin47:


tanin47:
It turns out we need to set CircleCI to share coveralls' token with a fork. This means a fork can access our environment variables. But it's not a concern because coveralls' token is the only env variable.

Issue #346, opened 2017-06-13
We should change --disable-gclient to --skip-deps


tanin47:
`install-build-deps.sh` is slow. `--skip-deps` will skip all pre-build steps (e.g. `gclient`)

This might need more discussion.

tanin47:
The main rational is that, once we run these deps, we don't need to re-run them in order to reproduce in a subsequential run.

`--skip-deps` will skip any command that fits the above behaviour.

PR #363, opened 2017-06-14
Add --skip-deps

tanin47:
- This will skip all dependency installation. This allows faster iteration when user runs the tool on the same testcase multiple times.
- `pre_build_steps` is separated into 3 methods: `gclient_sync`, `gclient_runhooks`, and `install_deps`
- `setup_args_gn` is separated into 2 methods: `setup_args_gn`, and `gn_gen`.

This solves #346.


README.md
```diff
@@ -75,17 +64,19 @@ optional arguments:
   -j GOMA_THREADS, --goma-threads GOMA_THREADS
                         Manually specify the number of concurrent jobs for a
                         ninja build.
+  -l GOMA_LOAD, --goma-load GOMA_LOAD
```
Dor1s: Should the text be aligned in the same way as other lines are?

tool/clusterfuzz/binary_providers.py
```diff
@@ -270,38 +292,31 @@ def setup_gn_goma_params(self, gn_args):
...
-      gn_args = self.gn_args
-    else:
-      args_gn_downloaded_build_path = os.path.join(
-          self.build_dir_name(), 'args.gn')
-      with open(args_gn_downloaded_build_path, 'r') as f:
-        gn_args = f.read()
+    gn_args = read_gn_args(
+        self.gn_args,
+        downloaded_args_gn_path = os.path.join(
+            self.build_dir_name(), 'args.gn'))
```
Dor1s: Would it make sense to put `'args.gn'` into a global constant?

tool/clusterfuzz/binary_providers.py
```diff
@@ -339,19 +375,20 @@ def get_goma_load(self):
...
+    self.gn_gen()
 
     common.execute(
         'ninja',
-        "-w 'dupbuild=err' -C %s -j %i -l %i %s" % (
-            self.build_directory, goma_cores, goma_load, self.target),
-        self.source_directory, capture_output=False,
+        ("-w 'dupbuild=err' -C {build_dir} -j {goma_cores} -l {goma_load} "
+         '{target}'.format(
+            build_dir=self.build_directory,
```
Dor1s: Should it be aligned by 2 spaces more? And the following 3 lines as well

tool/tests/clusterfuzz/binary_providers_test.py
```diff
@@ -653,89 +584,79 @@ def test_get_binary_path(self):
...
     helpers.patch(self, [
-        'clusterfuzz.common.execute',
-        'clusterfuzz.binary_providers.sha_from_revision',
-        'clusterfuzz.binary_providers.V8Builder.setup_gn_args'])
+        'clusterfuzz.binary_providers.gclient_runhooks_msan',
+        'clusterfuzz.binary_providers.sha_from_revision'])
 
     testcase = mock.Mock(id=12345, build_url='', revision=4567,
-                         gn_args='msan_track_origins=2\n')
+                         gn_args={'msan_track_origins': '4'})
```
Dor1s: Is `msan_track_origins=4` supported? I thought it can be `0, 1, 2`: https://www.chromium.org/developers/testing/memorysanitizer#TOC-Origin-tracking

README.md
```diff
@@ -75,17 +64,19 @@ optional arguments:
...
   -i ITERATIONS, --iterations ITERATIONS
                         Specify the number of times to attempt reproduction.
   -dx, --disable-xvfb   Disable running testcases in a virtual frame buffer.
   --target-args TARGET_ARGS
                         Additional arguments for the target (e.g. chrome).
   --edit-mode           Edit args.gn before building and target arguments
                         before running.
-  --disable-gclient     Disable running gclient commands (e.g. sync,
-                        runhooks).
+  --skip-deps           Skip installing dependencies: gclient sync, gclient
```
Dor1s: I personally would prefer disabling `install-build-deps.sh` separately from `gclient` commands, as you correctly noticed below that *"This is separated from install_deps because it is needed in every build"*:
* I believe that `install-build-deps.sh` can be disabled by default, since it is not needed in most of the cases.
* `gclient sync` and `gclient runhooks` should be ON by default, but it should be possible to disable them manually.

Feel free to ignore this comment if I'm wrong, I may be slightly out of context :)

tool/tests/clusterfuzz/binary_providers_test.py
tanin47: It's merely for unit testing. The value shouldn't matter here.

tool/clusterfuzz/binary_providers.py
tanin47: Good catch. I'll make a constant out of it.

README.md
tanin47: > "This is separated from install_deps because it is needed in every build"

I mean for every type of build.

I think the `--skip-deps` is for every command that isn't necessary (after the first reproduction) for reproduction when only the code is changed.

It's a slippery slope. Previously, we offered `--disable-gclient`. Then, user wants to disable `install-build-deps.sh`. I think user would want to disable `download_gold_plugin.py` as well. Therefore, we simplify it to running either all of them or none at all (as opposed to offering flag for disabling each of them).

In any case, the new structure is suitable for offering disabling flag for each command as well, if we change our mind.

> I believe that install-build-deps.sh can be disabled by default, since it is not needed in most of the cases.

Yes, it isn't clear here. But we only run `install-build-deps.sh` for the 32-bit job type.

README.md
tanin47: I think it's aligned... This is copied directly from `reproduce -h`.

tool/clusterfuzz/binary_providers.py
tanin47: Done. Thanks.

tanin47:
This is ready for a review now. cc @Dor1s

PR #368, opened 2017-06-16
Fix CI reading the log file.

tanin47:
The problem is `f.seek()` was going back too far. And `pyfakefs` doesn't simulate the real behaviour.


Dor1s:
LGTM

PR #372, opened 2017-06-18
Simplify CI loop and print when skipping a testcase

tanin47:
- Print when skipping a testcase because its job type is not supported. I've found that, once a CI runs for a long time, it'll exhaust *all* supported testcases. Then, it'll keep skipping the unsupported ones.
- I simplified the CI a bit. I don't think we need the LRU cache since we'll never use all the memory by storing integers.
- Retry the testcase when the failure is `MinimizationNotFinishedError`.


ci/continuous_integration/tests/daemon/main_test.py
```diff
@@ -217,11 +228,9 @@ def test_get_testcase(self):
         mock.call(
             'https://clusterfuzz.com/v2/testcases/load',
             headers={'Authorization': 'Bearer xyzabc'},
-            json={'page': 1, 'reproducible': 'yes'}),
-        mock.call(
-            'https://clusterfuzz.com/v2/testcases/load',
-            headers={'Authorization': 'Bearer xyzabc'},
-            json={'page': 2, 'reproducible': 'yes'})
+            json={'page': page, 'reproducible': 'yes',
+                  'q': 'platform:linux', 'open': 'yes'})
```
Dor1s: 'q' stands for 'Query'?

Dor1s: LGTM, but just a quick question to clarify

tanin47: Yes, this is because clusterfuzz.com accepts `q`.

clusterfuzz.com accepts `q` because it'd make the URL more beautiful, and it follows the standard (e.g. https://www.google.com/?q=test).

Issue #374, opened 2017-06-19
Improve the error message from clusterfuzz.com


tanin47:
Especially, when getting the testcase from clusterfuzz.

Right now we just show the JSON response to the user.

So, it's a bit hard to understand what's going on.

PR #376, opened 2017-06-19
Improve error message around invalid testcase

tanin47:
When it's 404, 400, or 401, we print the error message and suggestion explicitly.

This solves #374


error/error/error.py
```diff
@@ -58,21 +58,20 @@ def __init__(self):
         self.MESSAGE, self.EXIT_CODE)
 
 
-class ClusterfuzzAuthError(ExpectedException):
-  """An exception to deal with Clusterfuzz Authentication errors.
+class ClusterfuzzError(ExpectedException):
```
inferno-chromium: s/Clusterfuzz/ClusterFuzz everywhere in source

error/error/error.py
```diff
@@ -235,3 +234,32 @@ def __init__(self, question):
...
+  def __init__(self, testcase_id):
+    super(InvalidTestcaseIdError, self).__init__(
+        self.MESSAGE.format(testcase_id=str(testcase_id)), self.EXIT_CODE)
+
+
+class UnauthorizedError(ExpectedException):
+  """An exception when the user cannot access the testcase."""
+
+  MESSAGE = (
+      "You aren't allow to access the testcase ID ({testcase_id}). "
```
inferno-chromium: s/allow/allowed

error/error/error.py
tanin47: Done

error/error/error.py
tanin47: Done

PR #377, opened 2017-06-19
--current --skip-deps become critical, so we add them to the CI.

tanin47:
People are using `--current --skip-deps` intensely. We need to make sure they don't fail.


ci/continuous_integration/daemon/main.py
```diff
@@ -221,12 +224,14 @@ def reset_and_run_testcase(testcase_id, category, release):
   process.call('git clean -d -f -f', cwd=CHROMIUM_SRC)
 
   version = prepare_binary_and_get_version(release)
-  update_auth_header()
-  return_code = run_testcase(testcase_id)
-  logs = read_logs()
 
-  stackdriver_logging.send_run(
-      testcase_id, category, version, release, return_code, logs)
+  for opts in TESTED_OPTS:
```
tanin47: What happens here is:

1. Run `clusterfuzz reproduce 12345`
2. Then, run `clusterfuzz reproduce 12345 --current --skip-deps` immediately afterward. The build time should be zero here.

ci/continuous_integration/daemon/main.py
```diff
@@ -36,6 +36,9 @@
     error.MinimizationNotFinishedError.EXIT_CODE
 ])
 
+# The options that will be tested on the CI.
+TESTED_OPTS = ['', '--current --skip-deps']
```
inferno-chromium: s/OPTS/OPTIONS everywhere please.

inferno-chromium: s/TESTED/TEST sounds better.

ci/continuous_integration/tests/daemon/main_test.py
```diff
@@ -94,11 +94,12 @@ def test_not_save_testcase_id(self):
...
         error.MinimizationNotFinishedError.EXIT_CODE, None)
     self.assertEqual(
-        error.MinimizationNotFinishedError.EXIT_CODE, main.run_testcase(1234))
+        error.MinimizationNotFinishedError.EXIT_CODE,
+        main.run_testcase(1234, ''))
 
     self.assert_exact_calls(self.mock.call, [
         mock.call(
-            '/python-daemon-data/clusterfuzz reproduce 1234',
+            '/python-daemon-data/clusterfuzz reproduce 1234 ',
```
inferno-chromium: nit: we can fix this to not have whitespace at end.

PR #378, opened 2017-06-20
Some small fixes

tanin47:
- Small fixes according to the last PR.
- Add some more suggestions. The just-in-time-style suggestions seem to be effective.


error/error/error.py
```diff
@@ -110,7 +110,12 @@ def __init__(self):
 class JobTypeNotSupportedError(ExpectedException):
   """An exception raised when user tries to run an unsupported build type."""
 
-  MESSAGE = 'The job {job_type} is not yet supported by clusterfuzz tools.'
+  MESSAGE = (
+      'Unfortunately, the job {job_type} is not yet supported.'
+      'If you believe that the crash will occur on Linux as well, please go '
+      'to https://clusterfuzz.com/?noredirect=1#uploadusertestcase, upload '
```
inferno-chromium: https://clusterfuzz.com/uploadusertestcase might be a better link.

tanin47: That's true. Should have thought of that.

Issue #381, opened 2017-06-20
ninja's error seems to be removed from screen


tanin47:

tanin47:
This is because `ninja` doesn't print error to `stderr`.

tanin47:
This will be merged and released next week. Thanks @npm1 for reporting this.

PR #382, opened 2017-06-21
Parse ninja failure

tanin47:
`ninja` doesn't print err to `stderr`... so we have to parse the error ourselves from `stdout`. This solves #381.


tool/clusterfuzz/output_transformer.py
```diff
@@ -59,33 +59,72 @@ def flush(self):
     self.write('')
 
 
+def contain_failure(lines):
```
inferno-chromium: contains_failure ?

tool/clusterfuzz/output_transformer.py
```diff
@@ -59,33 +59,72 @@ def flush(self):
...
+        self.print_line(line)
+        self.previous_failed = False
+
+  def print_line(self, line):
+    """Print a single line."""
+    line_size = len(line)
+
+    if not self.previous_failed:
+      if line_size < self.previous_line_size:
+        line += ' ' * (self.previous_line_size - line_size)
```
inferno-chromium: This seems like unneeded complexity, just bolding the previous line should be good enough.

tool/clusterfuzz/output_transformer.py
```diff
@@ -59,33 +59,72 @@ def flush(self):
...
+      self.write('\b' * self.previous_line_size)
+
+    self.write(line)
+
+    self.previous_line_size = len(line)
 
   def flush(self):
     """Print the residue output."""
-    self.process('\n')
+    self.print_block(self.lines)
```
inferno-chromium: Will this confuse non-ninja text ? like stacktrace, etc.

tanin47: No, because this output_transformer is only used with the ninja command.

tool/clusterfuzz/output_transformer.py
tanin47: Done

tool/clusterfuzz/output_transformer.py
tanin47: We need to print ` ` (empty space) because we need to erase the previous line with space. Otherwise, if the previous line is longer than the current line, then the line will contain the current line and the later part of the previous line. For example:

Line 1: `aaaaaaaaaa`
Line 2: `bbbbb`

If we didn't print the empty spaces, then it'll become `bbbbbaaaaa`.

tool/clusterfuzz/output_transformer.py
tanin47: Here's how we use it:

```
    common.execute(
        'ninja',
        ("-w 'dupbuild=err' -C {build_dir} -j {goma_cores} -l {goma_load} "
         '{target}'.format(
             build_dir=self.build_directory,
             goma_cores=self.get_goma_cores(),
             goma_load=self.get_goma_load(),
             target=self.target)),
        self.source_directory,
        capture_output=False,
        stdout_transformer=output_transformer.Ninja())
```

`common.execute` allows us to inject `stdout_transformer`. Therefore, this transformer doesn't apply to other commands.

tanin47:
It's ready for another pass. Thank you!

PR #389, opened 2017-06-27
Reset all children in third_party/ manually.

tanin47:
In Chromium's source code, most of the third-party dirs are git-ignored. We cannot reset those third-party dirs to their clean states.

`git clean -ffddx` deletes all ignored files; we'd need to pull all of them again with `gclient sync`. Then, we'd hit the git's rate limit.

Therefore, we traverse through all third-party dirs and reset their state manually.


ci/continuous_integration/daemon/main.py
```diff
@@ -212,6 +212,23 @@ def read_logs(path=CLUSTERFUZZ_LOG_PATH):
...
 
+def clean_third_party():
+  """Clean third_party/ dir. The children of third_party/ are git-ignored.
+    Therefore, git-clean doesn't work. We can't do `git clean -x` either
+    because we would need to pull everything again and hit git's rate limit.
+    Therefore, we go into each child (which is a git repo) and run
+    clean the repo manually."""
+  for root, dirs, _ in os.walk(os.path.join(CHROMIUM_SRC, 'third_party')):
+    for name in dirs:
+      basename = os.path.basename(name)
```
Dor1s: Is it possible that `name` is not a basename already?

tanin47: It's likely not. Because this works in our tests. If it was a basename, the `parent` would be blank, and the test would fail.

PR #393, opened 2017-06-28
Update GOMA on every deployment

tanin47:
Apparently, we are using a super old GOMA on the CI. And this causes an issue on GOMA server. I've been told to update GOMA. Therefore, I think we should update goma on every deployment.


ci/playbook.yml
```diff
@@ -64,6 +64,12 @@
       args:
         chdir: ../
         executable: /bin/bash
+    # We expect your goma to be at ~/goma.
+    - name: Update the local goma at ~/goma
+      shell: python goma_ctl.py update
+      args:
+        chdir: ~/goma
```
inferno-chromium: Can we use absolute paths please , like /home/goma ?

tanin47: It depends on the username. On my machine, it'll be `/home/tanin/goma`.

Dor1s: Can we use `GOMA_DIR` env var? I think it is always set up if a developer uses goma.

tanin47: `GOMA_DIR` on my machine is empty. But that would be more bullet-proof. I'll change to `GOMA_DIR` and default to `~/goma`.

inferno-chromium: SGTM

Dor1s: Hm, you are right, `GOMA_DIR` is not necessary. The [public doc](https://www.chromium.org/developers/gn-build-configuration) doesn't say anything about that

> Googlers can use this for distributed builds. goma_dir is only required if you put the Goma tools in a non-standard place (the default place is ~/goma or C:\goma\goma-win).

Though the internal doc suggests to use it.

PR #398, opened 2017-06-29
Get username the right way

tanin47:
- Goobuntu guarantess the env `USER`.... but other ubuntus don't (e.g. ci)
- Increment the image version because I want to create a new image.


tool/clusterfuzz/stackdriver_logging.py
```diff
@@ -16,6 +16,7 @@
 import json
 import time
 import os
+import pwd
```
inferno-chromium: This wont work on windows. i checked stackoverflow and getpass.getuser() seems to work on both.

tanin47: Done. Thank you.

PR #399, opened 2017-06-29
[Not ready] Parallel test

tanin47:


tanin47:
It's ugly to parallelize the commands. We will not do it.

PR #401, opened 2017-07-01
Fix cleaning, testcase caching, and permissions

tanin47:
- Suddenly, our instances started requiring permissions on `userinfo-email` (for goma) and `storage-ro` (for downloading a build). I'm not sure why.
- Added the testcase caching back. Currently, we don't cache the failed testcases (including `MinimizationNotFinishedError`). After a while, CI only runs on the failed testcases because (1) we prefer running the newest testcases and (2) the unminimized testcases are often the newest ones. In other words, @everestmz was correct to use `py_lru_cache`, which I removed it 2 weeks ago (https://github.com/google/clusterfuzz-tools/pull/372/files#diff-e3b2bd4eea7d0036b3ffc0e382c21d65L33)
- Cleaning the chromium source code is super tricky. This is the third attempt. It seems to work.


ci/playbook.yml
```diff
@@ -147,7 +150,8 @@
         mode: 'u=rwx,g=rwx,o=rwx'
       with_items:
         - { src: '/etc/sv/python-daemon', dest: '/etc/service/python-daemon' }
-    - pause:
-        seconds: 3
-        prompt: Waiting 3 second for the server to acknowledge the CI service
     - shell: "sv restart python-daemon"
+      retries: 10
```
tanin47: Retrying is better than waiting.

Issue #404, opened 2017-07-01
New CI issue


tanin47:
```
2017-07-01 23:25:38,064 [DEBUG]: Error: Command 'git checkout --quiet 86219d8c6a73f95e694b4e1594e1a8a0a33613b6' returned non-zero exit status 128 in /home/clusterfuzz/chromium/src/buildtools/third_party/libunwind/trunk
2017-07-01 23:25:38,064 [DEBUG]: fatal: reference is not a tree: 86219d8c6a73f95e694b4e1594e1a8a0a33613b6
```

This happens when doing `gclient sync`. Traversing and resetting every sub repo is probably too aggressive. Need a new strategy.

tanin47:
Here's another error:

```
2017-07-02 00:00:19,592 [DEBUG]: ----------------------------------------
2017-07-02 00:00:19,592 [DEBUG]: Error: Command 'git checkout --quiet c48bb8b02c98ae49438e43aa1143a958784822a5' returned non-zero exit status 128 in /home/clusterfuzz/chromium/src/third_party/swiftshader/third_party/pnacl-subzero
2017-07-02 00:00:19,592 [DEBUG]: fatal: reference is not a tree: c48bb8b02c98ae49438e43aa1143a958784822a5
2017-07-02 00:00:19,592 [DEBUG]: 
```

Interestingly, the release-ci doesn't have this issue.

the release-candidate-ci performs `gclient sync` with `--no-history --shallow`. It seems we need to get rid of these flags.

tanin47:
It seems `--no-history --shallow` is the root cause. `gclient` tries to run `git checkout` on a specific sha. But that sha doesn't exist locally because we run `--no-history --shallow`. This sounds like an improvement on `gclient`; it should fetch if the sha doesn't exitst. I'll file a bug to the `gclient`'s owner.

tanin47:
I filed a bug to chrome-infra here: https://bugs.chromium.org/p/chromium/issues/detail?id=738706

PR #406, opened 2017-07-02
Remove --shallow --no-history from gclient sync

tanin47:
This solves #404


tanin47:
This becomes worse than I thought.

`git fetch --shallow` will default all later `git fetch` to shallow. We have to run `git fetch --unshallow` to fix this. This happened on my machine.

It's a good thing we haven't shipped this to user. Otherwise, it'd cause huge confusion.

The CI instances are already recreated.

PR #408, opened 2017-07-03
Try gclient sync --reset; add gitcookies to avoid rate-limit

tanin47:
The CI was deployed an hour ago. We'll let it run to see what happened.


tanin47:
This seems to work well. CI has been running without glitches for more than 48 hours.

The thing that really screwed us is `gclient sync --no-history --shallow`.

PR #413, opened 2017-07-07
Small fixes to make deployment better.

tanin47:
Fix here and there to make it easier to deploy to a different project.


ci/continuous_integration/daemon/main.py
```diff
@@ -209,6 +209,10 @@ def clean():
   """Clean the repo and all of the sub repos. This combination seems to
     work. Please be careful when changing it."""
   process.call('git clean -ffdd', cwd=CHROMIUM_SRC)
+  # We've encountered the case where `git clean` and `git reset --hard` don't
```
tanin47: This part will likely be changed again. For now, it seems to work...

tool/clusterfuzz/resources/supported_job_types.yml
```diff
@@ -8,6 +8,7 @@ presets: #Any parameters in a preset can be overridden in the actual definition
             reproducer: Base
         pdfium_chromium:
             preset: pdfium_standalone
+            source_name: chromium
```
tanin47: This is a bug.

inferno-chromium: Why ? Do you have a fix for this. I think this was there for building pdfium in chromium repo ?

tanin47: We have two of them. One is `pdfium_standalone` which builds from pdfium source. This is `pdfium_chromium` which builds from chromium source.

The bug is from the big refactoring a few days ago.

PR #415, opened 2017-07-08
Fix Pex unicode issue

tanin47:
If we didn't import `sys`, we would get the error:

```
File "/python-daemon/daemon.pex", line 2
SyntaxError: Non-ASCII character '\xe4' in file /python-daemon/daemon.pex on line 3, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```

I don't know why.


tanin47:
The error is in `daemon.pex` itself. It's not in any of our files. But I can't explain why adding `import sys` makes it work. I'll try putting the header in every python file and report back.

tanin47:
This is super weird. Adding one blank line also solves the problem.

tanin47:
No need to merge this. Somehow it works now. I don't understand what was happening.

Issue #390, opened 2017-06-27
Cache downloaded testcase for 2 hours.


tanin47:
We can use this https://pypi.python.org/pypi/pyfscache to avoid implement anything ourselves.

This speeds up the workflow when user runs the tool multiple times to fix a crash.

PR #417, opened 2017-07-11
Cache the downloaded testcase and the testcase JSON

tanin47:
This would reduce the run time by 10 seconds or so. It's helpful when user are iterating their code with our reproduce tool. This solves #390


tool/clusterfuzz/commands/reproduce.py
```diff
@@ -220,10 +222,18 @@ def warn_unreproducible_if_needed(current_testcase):
         common.BASH_YELLOW_MARKER))
 
 
+def clear_cache_if_needed(should_invalidate):
+  """Invalidate all caches."""
+  if not should_invalidate:
```
inferno-chromium: This condition should be in the caller, as there is no point of a function argument that just bails out on the first line.

tool/clusterfuzz/main.py
```diff
@@ -76,6 +76,11 @@ def execute(argv=None):
           'Build Chrome with full debug symbols by injecting '
           '`sanitizer_keep_symbols = true` and `is_debug = true` to args.gn. '
           'Ready to debug with GDB.'))
+  reproduce.add_argument(
```
inferno-chromium: Is there any usecase for this ? If not, lets not complicate options. The more the options, the more complicated is for the users to read through them. Lets just keep stuff that the users have asked for in past. Please remove any other unneeded as well.

tool/clusterfuzz/testcase.py
```diff
@@ -79,6 +83,31 @@ def get_environment_and_args(stacktrace_lines):
   return new_env, args
 
 
+def download_testcase_if_needed(url, dest_dir):
+  """Download a file into the dest_dir with caching. dest_dir must be safe to
+    be deleted."""
+  if (os.path.exists(dest_dir) and
+      (time.time() - os.stat(dest_dir).st_ctime) <= TESTCASE_CACHE_TTL):
+    return
+  print time.time()
```
inferno-chromium: Probably these prints need to be removed.

tool/clusterfuzz/main.py
tanin47: I suspect that it might come up. An example is, when the minimization recently finished, then user might want to invalidate the cache and get the new testcase JSON immediately.

We can hide it in the help with `argparse.SUPPRESS`. When user has the issue, we can tell them to invalidate the cache with `--invalidate-cache`.

tool/clusterfuzz/testcase.py
tanin47: Whoops, this is a mistake. They shouldn't have been here. Thanks for catching that.

tool/clusterfuzz/commands/reproduce.py
tanin47: I prefer the method to handle the condition to disable.

If we put the condition in the caller (`execute`), then two tests would be needed to cover this condition to achieve 100% coverage. A test for `execute` would also test other things that are not related to clearing cache or the `invalidate_cache` condition; this means it takes more effort to mock and etc.

tool/clusterfuzz/main.py
inferno-chromium: Yes, lets hide by default. We can add this in a seperate section in docs, with a link to these detailed options.

tool/clusterfuzz/commands/reproduce.py
inferno-chromium: ok

tool/clusterfuzz/main.py
tanin47: Second thought. If we will hide it, we might as well tell user to `rm -rf ~/.clusterfuzz/cache/testcases` instead. The code will be simpler as well. Thanks for the suggestion.

PR #418, opened 2017-07-11
Fix the cache bug

tanin47:


tool/clusterfuzz/common.py
```diff
@@ -66,10 +66,11 @@
 )
 
 
-# Configuring backoff retrying because sending a request to ClusterFuzz
-# might fail during a deployment.
+# Ensure the testcase dir exists.
```
inferno-chromium: Why is this in a global context, it will be run everytime anyone imports the common module. Should probably happen only one time in the main __init__ at start of reproducer script.

tanin47: Our `HTTP` is a global variable. Then, we changed it to a cached HTTP. So, that's why we need to ensure the parent dir of the cache exists.

I don't like this either. Let me come up with a better design.

tanin47:
- Moved the http object from global space to a method with memoization.
- Refactor `memoize` to support module functions. Ensure `memoize` doesn't support arguments.
- Found some methods that use `memoize` with arguments... so we have to refactor those methods

tool/clusterfuzz/common.py
```diff
@@ -66,26 +67,59 @@
...
-        # backoff_factor is 0.5. Therefore, the max wait time is 16s.
-        retry.Retry(
-            total=5, backoff_factor=0.5, status_forcelist=[500, 502, 503, 504]))
-)
-
-
-def post(*args, **kwargs):  # pragma: no cover
-  """Make a post request. This method is needed for mocking."""
-  return HTTP.post(*args, **kwargs)
+def memoize(is_instance=True):
```
inferno-chromium: Why do you need is_instance at all ? Why can't make auto-decision based on len(args), that seems cleaner. otherwise we can have bugs when we don't use is_instance properly like initialize with wrong boolean.

tanin47: Using `len(args)` is incidental. It wouldn't allow us to check if `memoize` was used correctly (e.g. `assert len(args) == 1` or `assert not args`).

With `is_instance`, it's difficult to make a mistake as there's only one possible mistake: user might wrap `def test(a)` with `memoize(is_instance=True)`.

tanin47:
We've switched to memoizing by arguments. It works with both instance methods and module methods.

PR #428, opened 2017-07-21
Reproduce Android

tanin47:
In order to support Android, here are the changes:
- Add a new reproducer for Android.
- The testcase supports the `files` field which specify the location and the content of necessary files on Android.
- Filtering logs from Android is copied from ClusterFuzz.

Symbolization hasn't been implemented yet. This works with: 5971973158207488


inferno-chromium:
Oliver, can you take a first pass at this review.

tanin47:
Gentle ping

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def adb(arg, **kwargs):
+  """Run adb with arg."""
+  return common.execute('adb', arg, cwd='.', **kwargs)
+
+
+def adb_shell(arg, **kwargs):
+  """Run adb shell with arg."""
+  return adb('shell "%s"' % arg, **kwargs)
```
oliverchang: we should probably be more careful with escaping characters here since the input here is a shell command?

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def write_content(path, content):
+  """Write content to path on Android."""
+  with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+    print tmp_file
```
oliverchang: Are these prints intended to be left here?

tool/clusterfuzz/reproducers.py
```diff
@@ -546,3 +566,67 @@ def reproduce_crash(self):
...
+    """Reproduce crash on Android."""
+    android.reset(self.get_package_name())
+    android.ensure_active()
+    android.clear_log()
+
+    ret_value, _ = android.adb_shell(
+        'am start -a android.intent.action.MAIN '
+        "-n %s/com.google.android.apps.chrome.Main '%s'" %
+        (self.get_package_name(), self.get_testcase_path()))
+    time.sleep(30)
```
oliverchang: maybe put this in a constant somewhere?

tool/clusterfuzz/android.py
tanin47: Oops. thanks for pointing it out.

tool/clusterfuzz/reproducers.py
tanin47: Done

tool/clusterfuzz/android.py
tanin47: Done.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+  logger.info(
+      common.colorize(
+          '\nWriting %s:\n%s\n', common.BASH_GREEN_MARKER),
+      path, content)
+
+  adb('push %s %s' % (tmp_file.name, path))
+  common.delete_if_exists(tmp_file.name)
+
+
+def file_exist(path):
```
inferno-chromium: This is just wrong, please see adb.fileExists

Some adb commands can emit text like Warning:, so we need to check return code.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+
+logger = logging.getLogger('clusterfuzz')
+
+
+def adb(arg, **kwargs):
+  """Run adb with arg."""
+  return common.execute('adb', arg, cwd='.', **kwargs)
+
+
+def adb_shell(arg, **kwargs):
```
inferno-chromium: Please use code from runADBShellCommand

  def _escape_specials(command):
    return command.replace('\\', '\\\\').replace('"', '\\"')

  if isinstance(cmd, list):
    cmd = ' '.join([str(i) for i in cmd])

  if cmd[0] not in ['"', "'"]:
    cmd = '"{}"'.format(_escape_specials(cmd))

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+
+def file_exist(path):
+  """Check if the path exists on Android."""
+  _, output = adb_shell('ls %s 2> /dev/null' % path, exit_on_error=False)
+  return bool(output.strip())
+
+
+def ensure_asan(source_dir_path, device_id):
+  """Ensures ASan is installed on Android."""
+  if file_exist('/system/lib/libclang_rt.asan-arm-android.so'):
```
inferno-chromium: This file is different for 32-bit and 64-bit asan. please don't assume it, also that file can be updated from time to time. so, running it everytime is fast, it will skip if the checksums of changes are different, so run it everytime.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+    logger.info(
+        'The device has ASAN. Skip installing ASAN.')
+    return
+
+  logger.info(
+      'Installing ASAN into the Android device. The device will be restarted.')
+
+  common.check_confirm(
+      'Are you sure you want to install ASAN on the Android device %s?' %
```
inferno-chromium: You can say like this testcase needs to install ASAN on your device.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+      device_id)
+
+  common.execute(
```
inferno-chromium: Are you running tools/clang/update.py at any step. this will make sure clang is updated locally.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+  """Convert Android's crash stack line into sanitizer_format."""
+  m_crash_state = re.match(
+      r'\s*#([0-9]+)\s+pc\s+([xX0-9a-fA-F]+)\s+(.+)', line)
+  if not m_crash_state:
+    return None
+
+  frame_no = int(m_crash_state.group(1))
+  frame_address = m_crash_state.group(2)
+  frame_binary = m_crash_state.group(3).strip()
+  # Ignore invalid frames, helps to prevent errors
```
inferno-chromium: new line before this line.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+  # while symbolizing.
+  if '<unknown>' in frame_binary:
+    return
+  # Normalize frame address.
```
inferno-chromium: new line before this line.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+    m_line = re.match('[^D]/([^:]+)[:] (.*)', line)
+    if not m_line:
+      continue
+
+    header = m_line.group(1)
+    content = m_line.group(2)
+
+    process_id, process_name = get_process_id_and_name(header)
+    if process_id != last_process_id:
+      filtered_output += '--------- %s (%d):\n' % (process_name, process_id)
```
inferno-chromium: Isn't this to be done after process android and chrome stack lines. ? check original code in clusterfuzz.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+  adb_shell(
+      'input keyevent KEYCODE_POWER', print_command=False, print_output=False)
+
+  # Unlock.
+  adb_shell(
+      'input keyevent KEYCODE_MENU', print_command=False, print_output=False)
+  time.sleep(1)
+
+
+def set_content(table, key, value):
```
inferno-chromium: s/set_content/set_content_setting as a less confusing name.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+      redirect_stderr_to_stdout=True,
+      stdout_transformer=output_transformer.Identity())
+  return output
+
+
+def kill(package_name):
+  """Kills the process with the package name."""
+  adb_shell('am force-stop %s' % package_name, exit_on_error=False)
+
+  # Click OK to close 'Unfortunately, Chrome has stopped'.
```
inferno-chromium: Add a FIXME, this is very hacky.

tool/clusterfuzz/android.py
tanin47: Yes, the base class of binary providers run `tools/clang/update.py` for every build.

tool/clusterfuzz/android.py
tanin47: Done

tool/clusterfuzz/android.py
tanin47: Done

error/error/error.py
```diff
@@ -294,3 +294,40 @@ def __init__(self, count, crash_signatures):
...
+
+  def __init__(self):
+    super(WaitForAndroidAfterInstallingAsanError, self).__init__(
+        message=self.MESSAGE, exit_code=self.EXIT_CODE)
+
+
+class NoAndroidDeviceIdError(ExpectedException):
+  """An exception is raised after installing ASAN on Android"""
+
+  MESSAGE = 'Please set the target Android device ID as the env {env_name}.'
```
tanin47: The message is only for this type of error. Every error has its own `MESSAGE`.

tool/clusterfuzz/android.py
tanin47: In ClusterFuzz, it's processed *after*. But its result is added *before*. Other than that, their processing are independent (e.g. one uses header, the other uses body).

tool/clusterfuzz/android.py
tanin47: Done.

tool/clusterfuzz/android.py
tanin47: Done

tool/clusterfuzz/android.py
tanin47: Done.

tool/clusterfuzz/android.py
tanin47: Done.

tool/clusterfuzz/android.py
tanin47: Anecdotally, `test -f` doesn't seem to work well. For example:

```
$ adb shell "ls /system/lib/libvixl.so"
/system/lib/libvixl.so
$ adb shell "test -f /system/lib/libvixl.so; echo $?"
0
$ adb shell "test -f /system/lib/libvixlssss.so; echo $?"
0
```

If there's a warning, how about we check if the `path` is in `output` (after splitting by lines)?

tool/clusterfuzz/android.py
tanin47: Done.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,244 @@
...
+  with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+    tmp_file.write(content)
+
+  logger.info(
+      common.colorize(
+          '\nWriting %s:\n%s\n', common.BASH_GREEN_MARKER),
+      path, content)
+
+  adb('push %s %s' % (tmp_file.name, path))
+  common.delete_if_exists(tmp_file.name)
```
tanin47: Done. I don't think we need remount here because we only push to /sdcard folder.

tanin47:
Thank you for the comments. It's ready for another pass.

error/error/error.py
inferno-chromium: Sorry i meant for {env_name}, that shouldn't change.

tool/clusterfuzz/android.py
inferno-chromium: Lets discuss this tmrw, i definitely remember it worked before.

error/error/error.py
```diff
@@ -294,3 +294,40 @@ def __init__(self, count, crash_signatures):
...
+    super(GdbNotSupportedOnAndroidError, self).__init__(
+        message=self.MESSAGE, exit_code=self.EXIT_CODE)
+
+
+class WaitForAndroidAfterInstallingAsanError(ExpectedException):
+  """An exception is raised after installing ASAN on Android"""
+
+  MESSAGE = (
+      'Please wait until the Android device is rebooted and run the tool '
+      "again. If there's permission error, try `adb remount` and run the "
```
inferno-chromium: s/try `adb remount`/make sure that `adb remount` succeeds

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,253 @@
...
+
+SCREEN_LOCK_SEARCH_STRING = 'mShowingLockscreen=true'
+ASAN_DEVICE_SETUP_SUCCESS_SEARCH_STRING = (
+    'Please wait until the device restarts')
+
+logger = logging.getLogger('clusterfuzz')
+
+
+def adb(command, **kwargs):
+  """Run adb with arg."""
```
inferno-chromium: fix description for both function to have command.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,253 @@
...
+      common.colorize(
+          '\nWriting %s:\n%s\n', common.BASH_GREEN_MARKER),
+      path, content)
+
+  adb('push %s %s' % (tmp_file.name, path))
+  adb_shell('chmod 0644 %s' % path)
+  common.delete_if_exists(tmp_file.name)
+
+
+def file_exist(path):
```
inferno-chromium: s/file_exist/file_exists ?

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,253 @@
...
+  """Check if the path exists on Android."""
+  _, output = adb_shell('ls %s 2> /dev/null' % path, exit_on_error=False)
```
inferno-chromium: Are we assuming that this is an absolute_path, if yes, lets rename vars.

tool/clusterfuzz/android.py
```diff
@@ -0,0 +1,253 @@
...
+              'lib', 'clang', '*', 'lib', 'linux'),
+          device_id),
+      env={'ADB_PATH': common.check_binary('adb', cwd='.')},
+      redirect_stderr_to_stdout=True,
+      cwd='.')
+
+  # tool/clusterfuzz/resources/asan_device_setup.sh prints the below string
+  # on screen. We've decided to do this hack in order to avoid modifying
+  # asan_device_setup.sh which is copied from Chromium's src.
+  if ASAN_DEVICE_SETUP_SUCCESS_SEARCH_STRING in output:
```
inferno-chromium: This is very very hacky, doesn't script return exit code. there are 2 cases, one where it changes stuff on device and exits with success, 2nd case where it does not need to modify anything and returns with success. otherwise it fails.

tool/clusterfuzz/reproducers.py
```diff
@@ -546,3 +566,67 @@ def reproduce_crash(self):
...
+
+    if not device_id:
+      raise error.NoAndroidDeviceIdError(ANDROID_SERIAL_ENV)
+
+    return device_id
+
+  @common.memoize
+  def get_package_name(self):
+    """Get package name."""
+    return 'org.chromium.chrome'
```
inferno-chromium: This seems wrong, we use com.google.android.apps.chrome in job type.

tool/clusterfuzz/reproducers.py
```diff
@@ -546,3 +566,67 @@ def reproduce_crash(self):
...
+    """Get the android device."""
+    device_id = os.environ.get(ANDROID_SERIAL_ENV)
+
+    if not device_id:
+      raise error.NoAndroidDeviceIdError(ANDROID_SERIAL_ENV)
+
+    return device_id
+
+  @common.memoize
+  def get_package_name(self):
```
inferno-chromium: I think like get_package_name,  line 614-615, 626-627, etc are chrome specific stuff which can go in AndroidChromeReproducer inherited class. Or just change this class name.

tool/clusterfuzz/reproducers.py
```diff
@@ -546,3 +566,67 @@ def reproduce_crash(self):
...
+
+  def reproduce_crash(self):
+    """Reproduce crash on Android."""
+    android.reset(self.get_package_name())
+    android.ensure_active()
+    android.clear_log()
+
+    ret_value, _ = android.adb_shell(
+        'am start -a android.intent.action.MAIN '
+        "-n %s/com.google.android.apps.chrome.Main '%s'" %
```
inferno-chromium: This wont work for webview shell job type, why do we need to hardcode this, why don't we get this from job environment

tool/clusterfuzz/reproducers.py
```diff
@@ -546,3 +566,67 @@ def reproduce_crash(self):
...
+    android.ensure_asan(
+        source_dir_path=self.binary_provider.get_source_dir_path(),
+        device_id=device_id)
+
+    super(AndroidReproducer, self).pre_build_steps()
+
+    for path, content in self.testcase.files.iteritems():
+      android.write_content(path, content)
+    android.write_content(
+        '/data/local/chrome-command-line', 'chrome %s' % self.args)
```
inferno-chromium: Good to get this from job environment COMMAND_LINE_PATH ?

error/error/error.py
tanin47: Done.

tool/clusterfuzz/android.py
tanin47: Done.

tool/clusterfuzz/android.py
tanin47: Ok, funnily enough, since we remove the asan check (e.g. checking if `/system/lib/...asan...` exist, we don't actually use this method anymore..

tool/clusterfuzz/android.py
tanin47: The method is removed... we don't use it anymore after removing the asan check.

tool/clusterfuzz/android.py
tanin47: The script succeeding (exiting with 0) means it either installs ASAN successfully or skips the installation.

Sadly, the script doesn't exit with different codes (https://github.com/google/clusterfuzz-tools/pull/428/files#diff-d224b682b1ec5006b23abcefaf30b29cR459). Here's the relevant code:

```
+    echo '>> Please wait until the device restarts'
 +else
 +    echo '>> Device is up to date'
 +fi
```

We need to distinguish between these 2 cases because we need stop if ASAN installation previously occurred.

inferno-chromium: Isn't "Device is up to date" also a success.

tool/clusterfuzz/reproducers.py
tanin47: I switched to reading from the stacktrace instead.

tool/clusterfuzz/reproducers.py
tanin47: I think this is because we aren't building with Clank. We build from Chromium directly. Maybe that's a mistake. I'll change it to Clank.

tanin47: Actually, this involves changing binary_providers. Let's do that in a separate CL.

tool/clusterfuzz/reproducers.py
tanin47: I'll fix this in a separate CL. We'll switch to Clank and get the command from the stacktrace.

Currently, the package name when building from Chrome is different. That's why this is hardcoded.

tool/clusterfuzz/android.py
tanin47: I've changed the name to `ASAN_BEING_INSTALLED_SEARCH_STRING`.

tool/clusterfuzz/reproducers.py
tanin47: Done.

error/error/error.py
tanin47: I see what you mean now. The main issue here is that I wouldn't like to hardcode the string `ANDROID_DEVICE_ID`. It's defined in `reproducers.ANDROID_DEVICE_ID_ENV`; it's also a bit better to let consumer of this class define the variable name.

tanin47:
I've fixed and addressed all the comments. It's ready for another pass. Thank you.

error/error/error.py
inferno-chromium: Ok.

Issue #429, opened 2017-07-27
The dirty repo is still happening.


tanin47:
```
2017-07-26 22:57:24,846 [INFO]: Running: gclient sync
2017-07-26 22:57:24,849 [DEBUG]: ---------------------------------------
2017-07-26 22:57:40,226 [DEBUG]: Killing pid=41967 with 15
2017-07-26 22:57:40,226 [DEBUG]: 
2017-07-26 22:57:40,226 [DEBUG]: src/testing/gmock (ERROR)
2017-07-26 22:57:40,226 [DEBUG]: ----------------------------------------
2017-07-26 22:57:40,226 [DEBUG]: [0:00:00] Started.
2017-07-26 22:57:40,226 [DEBUG]: ----------------------------------------
2017-07-26 22:57:40,226 [DEBUG]: Error: 9> 
2017-07-26 22:57:40,226 [DEBUG]: 9> ____ src/testing/gmock at 0421b6f358139f02e102c9c332ce19a33faf75be
2017-07-26 22:57:40,226 [DEBUG]: 9> 	You have unstaged changes.
2017-07-26 22:57:40,226 [DEBUG]: 9> 	Please commit, stash, or reset.
2017-07-26 22:57:40,227 [DEBUG]: ---------------------------------------
2017-07-26 22:57:40,227 [DEBUG]: | Return code is non-zero (1).
2017-07-26 22:57:40,227 [DEBUG]: | Exit.
2017-07-26 22:57:40,367 [INFO]: CommandFailedError: `gclient sync` failed with the return code 1.
```

This dirty repo thing is still happening even with `gclient sync --force --reset`. I think we are gonna do a nuclear option:

```
rm -rf testing 
rm -rf third_party
git checkout HEAD testing
git checkout HEAD third_party
```

Note that the CI has been running fine for almost a week. It starts to get this error after hitting old testcases (> 5 months ago).

tanin47:
I think I might know how it happens.

For a really old testcase, `gclient sync` will just fail. An example is:

```
2017-07-27_22:14:07.20683 src/third_party/skia (ERROR)
2017-07-27_22:14:07.20683 ----------------------------------------
2017-07-27_22:14:07.20684 [0:00:03] Started.
2017-07-27_22:14:07.20684 [0:00:04] Up-to-date; skipping checkout.
2017-07-27_22:14:07.20684 _____ src/third_party/skia at 99d4721171f9b8f23cd907ef3da938c4c5579ae9
[...]

PR #431, opened 2017-07-27
Nuke the folder to reset repo.

tanin47:
We are nuking the folder manually because the issue in #429.


tanin47:
I added an explanation why `gclient sync` is ineffective in cleaning the repo in #429.

Issue #434, opened 2017-07-28
Drop `--check` from `gn gen`.


tanin47:
It causes an exception that we don't care about. for example:

```
2017-07-28 22:43:47,888 [INFO]: Running: gn gen --check /home/predator/chromium/src/out/clusterfuzz_5112446519083008
2017-07-28 22:43:47,894 [DEBUG]: ---------------------------------------
2017-07-28 22:43:48,294 [DEBUG]: ERROR at //v8/BUILD.gn:8:1: Can't load input file.
2017-07-28 22:43:48,295 [DEBUG]: import("//build/config/host_byteorder.gni")
2017-07-28 22:43:48,295 [DEBUG]: ^-----------------------------------------
2017-07-28 22:43:48,295 [DEBUG]: Unable to load:
2017-07-28 22:43:48,295 [DEBUG]:   /home/predator/chromium/src/build/config/host_byteorder.gni
2017-07-28 22:43:48,295 [DEBUG]: I also checked in the secondary tree for:
2017-07-28 22:43:48,295 [DEBUG]:   /home/predator/chromium/src/build/secondary/build/config/host_byteorder.gni
2017-07-28 22:43:48,295 [DEBUG]: See //BUILD.gn:83:15: which caused the file to be included.
2017-07-28 22:43:48,295 [DEBUG]:     deps += [ "//v8:gn_all" ]
2017-07-28 22:43:48,304 [DEBUG]:               ^------------
2017-07-28 22:43:48,305 [DEBUG]: Killing pid=7703 with 15
2017-07-28 22:43:48,305 [DEBUG]: ---------------------------------------
2017-07-28 22:43:48,305 [DEBUG]: | Return code is non-zero (1).
2017-07-28 22:43:48,305 [DEBUG]: | Exit.
```

tanin47:
Chromium documentation doesn't suggest `--check`. Only buildbot uses it.

Issue #433, opened 2017-07-28
Setting ASAN_OPTIONS while building with ninja might cause a crash


tanin47:
We have found one case where the user sets `ASAN_OPTIONS="symbolize=1 detect_leaks=1"` and runs the reproduce tool. The reproduce tool fails on `ninja` with an ASAN stacktrace.

This is because, when `is_asan = true`, everything is built with ASAN including `genperf`, and `genperf` seems to be have an ASAN bug (https://bugs.chromium.org/p/chromium/issues/detail?id=382380).

@Dor1s suggests that we simply remove `ASAN_OPTIONS` and every other sanitizers' envs when running `ninja`.

tanin47:
I can reproduce this on my machine as well.

inferno-chromium:
Yes correct, just unset *SAN_OPTIONS, others like MSAN_OPTIONS etc can have same consequences.

tanin47:
Special thanks to @Dor1s for helping. I'll turn off every `*SAN_OPTIONS`.

Dor1s:
Found this list on CF side, let's check all of those + probably LSAN:

```python
SUPPORTED_MEMORY_TOOLS_FOR_OPTIONS = ['ASAN', 'CFI', 'MSAN', 'TSAN', 'UBSAN']
```

PR #436, opened 2017-07-30
Drop --check; Unset *SAN_OPTIONS when using ninja

tanin47:
This solves  #433 and #434.


tool/clusterfuzz/binary_providers.py
```diff
@@ -37,7 +37,14 @@
     '{cmd} in {source_dir}?')
 ARGS_GN_FILENAME = 'args.gn'
 GOMA_DIR = os.path.expanduser(os.path.join('~', 'goma'))
-
+NINJA_ENV = {
```
inferno-chromium: Why are we just unsetting just for ninja ? I think we should just unset at start of reproduce tool run, since later when we try to reproduce crash, any already set *_OPTIONS might also cause unexpected results.

tanin47: Good point. Will fix it.

PR #437, opened 2017-07-30
Switch to Clankium

tanin47:
- `gclient sync` uses `clankium/src/clank` as the anchor. In other words, it uses `clankium/src/clank/DEPS` to sync all other repos' SHAs including `clankium/src`'s SHA.
- Get package name and main class from the stacktrace.


tool/clusterfuzz/binary_providers.py
```diff
@@ -72,6 +73,24 @@ def get_pdfium_sha(chromium_sha):
...
 
 
+def get_clank_sha(revision_url):
+  """Get Clank SHA."""
+  tmp_file = tempfile.NamedTemporaryFile(delete=False)
+  tmp_file.close()
+
+  common.gsutil('cp %s %s' % (revision_url, tmp_file.name), cwd='.')
+
+  with open(tmp_file.name, 'r') as f:
```
inferno-chromium: Don't like one char variable names.

s/f/file_handle
s/m/match below

tool/clusterfuzz/binary_providers.py
```diff
@@ -72,6 +73,24 @@ def get_pdfium_sha(chromium_sha):
...
+    body = f.read()
+  common.delete_if_exists(tmp_file.name)
+
+  m = re.search('"clank_revision": "([^"]+)"', body)
```
inferno-chromium: instead of [^"]+ better to use hash syntax [a-fA-F0-9]+

tool/clusterfuzz/testcase.py
```diff
@@ -155,6 +155,21 @@ def get_file_contents_for_android(environment_sections):
...
 
 
+def get_package_and_main_class_names(stacktrace_lines):
+  """Get package and main class names."""
+  for line in stacktrace_lines:
+    content = line['content'].strip()
+    if not content.startswith('shell am start'):
+      continue
+
+    m = re.match("shell am start -a [^ ]+ -n ([^/]+)/([^ ]+) .+", content)
```
inferno-chromium: Instead of [^ ]+ we should use [^\s]+, covers all kind of space including tabs.

tool/clusterfuzz/binary_providers.py
tanin47: Done

tool/clusterfuzz/testcase.py
tanin47: Done

tool/clusterfuzz/binary_providers.py
tanin47: Done

PR #439, opened 2017-07-31
Improve getting binary path

tanin47:
- Previously, we downloaded a zip file, extracted it, and expected the folder name to be the same as the  zip file. It doesn't work with Android because android's zip file contains all files on the root.
- Previously, we downloaded a zip file, extracted it, and chmoded the binary. This doesn't make sense. We should chmod it when we want to use it (Because, at that time, the binary name is available e.g. for libfuzzer case).
- Previously, `DownloadedBinary` was its own class. This doesn't make sense because, for libfuzzer, we need a mechanism to get the binary name from stacktrace, and the mechanism is in `LibfuzzerAndAflBuilder`. Constructing a multiple-inherited class at run-time avoids code redundancy. I hope to use this approach with Msan in the future.


tool/clusterfuzz/binary_providers.py
```diff
@@ -311,12 +329,7 @@ def get_build_dir_path(self):
     """Returns the location of the correct build to use for reproduction."""
     path = os.path.join(
         common.CLUSTERFUZZ_BUILDS_DIR, '%s_downloaded_build' % self.testcase.id)
-    # TODO(tanin): This doesn't work with libfuzzer or afl because there's no
```
tanin47: This is what we fix here.

tool/clusterfuzz/binary_providers.py
```diff
@@ -179,7 +181,19 @@ def serialize_gn_args(args_hash):
...
+def find_file(target_filename, parent_dir):
+  """Return the full path under parent_dir. In Chrome, the binary is
+    inside a sub-directory. But, in Android, the binary is in the parent dir."""
+  for root, _, files in os.walk(parent_dir):
+    for filename in files:
+      if filename == target_filename:
+        return os.path.join(root, filename)
+
+  raise Exception(
+      'Cannot find the file name (%s) in %s.' % (target_filename, parent_dir))
```
inferno-chromium: s/Cannot find the file name (%s) in %s/Cannot find file name (%s) in directory (%s) sounds a little better.

tool/clusterfuzz/binary_providers.py
```diff
@@ -194,18 +208,17 @@ def download_build_if_needed(dest, url, binary_name):
...
+  tmp_dir_path = tempfile.mkdtemp(dir=common.CLUSTERFUZZ_CACHE_DIR)
+  common.execute('unzip', '-q %s -d %s' % (saved_file, tmp_dir_path), cwd='.')
 
-  logger.info('Cleaning up...')
-  os.remove(saved_file)
-  os.rename(os.path.join(
-      common.CLUSTERFUZZ_BUILDS_DIR, os.path.splitext(filename)[0]), dest)
+  # args.gn is guaranteed to be in the wanted folder. In Chrome, it's under a
+  # sub-directory. In Android, it's in the top dir.
+  root_dir = find_file('args.gn', tmp_dir_path)
```
inferno-chromium: This seems wrong. return value is not root_dir, it is args_gn_file_path and also next command looks weird, why does it need to use os.path.dirname ?

tool/clusterfuzz/binary_providers.py
```diff
@@ -194,18 +208,17 @@ def download_build_if_needed(dest, url, binary_name):
   filename = os.path.basename(gsutil_path)
   saved_file = os.path.join(common.CLUSTERFUZZ_CACHE_DIR, filename)
 
-  common.execute(
-      'unzip', '-q %s -d %s' % (saved_file, common.CLUSTERFUZZ_BUILDS_DIR),
-      cwd=common.CLUSTERFUZZ_DIR)
+  tmp_dir_path = tempfile.mkdtemp(dir=common.CLUSTERFUZZ_CACHE_DIR)
```
inferno-chromium: Why do we create tmp directory inside cache directory ? Shouldn't we have another top level tmp directory ? Seems weird by naming.

tool/clusterfuzz/binary_providers.py
```diff
@@ -290,8 +303,13 @@ def __init__(self, testcase, definition, options):
     self.definition = definition
     self.options = options
 
+  @common.memoize
   def get_binary_path(self):
-    return '%s/%s' % (self.get_build_dir_path(), self.get_binary_name())
+    """Return binary path and ensure it's executable."""
+    path = '%s/%s' % (self.get_build_dir_path(), self.get_binary_name())
```
inferno-chromium: use os.path.join for cross-platform.

tool/clusterfuzz/binary_providers.py
tanin47: Fixed in the #437

tool/clusterfuzz/binary_providers.py
tanin47: Fixed in the #437

tool/clusterfuzz/binary_providers.py
tanin47: Fixed in the #437

tool/clusterfuzz/binary_providers.py
tanin47: Fixed in the #437ixed in the #437

PR #441, opened 2017-07-31
Fix Android WebView

tanin47:
- Previously, we only support *one* target name. But Android's WebView needs to build *two* targets. So, we now support multiple target names.
- For Android WebView, we need to install `SystemWebViewGoogle.apk` and `SystemWebViewShell.apk`.
- Android distinguishes between `testcase_path` and `testcase_url` (e.g. `file://`). So, we refactor that path.
- When installing an apk on Android, it doesn't report failure... so we detect that manually.
- When using a downloaded build, the `libclang_rt.asan-arm-android.so` now comes from the downloaded build.

This reproduces 5603920566288384


tool/clusterfuzz/reproducers.py
```diff
@@ -618,12 +641,30 @@ def reproduce_crash(self):
...
+
+class AndroidWebViewReproducer(AndroidChromeReproducer):
+  """Reproduce an android webview job type."""
+
+  def install(self):
+    """Install webview."""
+    android.adb_shell(
+        'setprop persist.sys.webview.vmsize %s' % SYSTEM_WEBVIEW_VMSIZE_BYTES)
+    android.adb_shell('stop')
+    android.adb_shell('rm -rf %s' % ' '.join(SYSTEM_WEBVIEW_DIRS))
```
inferno-chromium: This won't work with adb remount ? where do you do that. original code does that. also original code does adb root to make sure root is there in adb.stopShell and adb.startShell, maybe you can create helpers that run adb root specifically.

tanin47: We called `adb.root` in `pre_build_step`, so I didn't repeat it here. I'll add remount when we do adb root.

PR #443, opened 2017-08-02
Android symbolization

tanin47:
We correct the path in the stacktrace. If the lib doesn't exist on the machine, we pull the lib from the device.

This works with both download builds and locally-built builds.


tanin47:
Ping. Please take a look. Thank you.

tool/clusterfuzz/android.py
```diff
@@ -158,6 +159,43 @@ def filter_log(content):
...
+    full path."""
+  filename = os.path.basename(binary_path)
+
+  for path in search_paths + [lib_tmp_dir_path]:
+    full_path = os.path.join(path, filename)
+    if os.path.exists(full_path):
+      return full_path
+
+  adb('pull %s %s' % (binary_path, lib_tmp_dir_path))
+  return os.path.join(lib_tmp_dir_path, filename)
```
inferno-chromium: test for path exists on local, what if lib is not on device.

tanin47: I'll raise an  exception on it. thanks for pointing out.

PR #457, opened 2017-08-11
Use our own python

### LINKED ISSUE #456: Make Pants work on Mac

tanin47:
This fixes #456.


tanin47:
This is misunderstood. It turns out that `pip` on vm60-m84 was broken. The error is `TypeError: __call__() takes exactly 2 arguments (1 given)`.

I fixed it with `python -m pip install -U setuptools`.

PR #468, opened 2017-08-14
Report enable_debug

tanin47:
Currently, if the user uses `enable_debug`, we might report it as reproducing successfully (depending on `gdb`'s return code). Therefore, we surface the enable debug flag in the log message, so that we are aware of it.


tool/tests/clusterfuzz/stackdriver_logging_test.py
```diff
@@ -43,13 +43,14 @@ def test_send_stacktrace(self):
...
-              'disable_goma': True}
+              'disable_goma': True,
+              'enable_debug': True}
     stackdriver_logging.send_log(params, 'Stacktrace')
 
     params['user'] = 'name'
     params['sessionId'] = 'user:1234:sessionid'
     params['message'] = (
-        'name successfully finished (reproduce, 123456, True).\n'
+        'name successfully finished (reproduce, 123456, True, debug).\n'
```
inferno-chromium: Why do we use True for current, we should just do 'current' just like 'debug'.

tanin47: Good point. I'll change it.

PR #471, opened 2017-08-15
Fix small things

tanin47:
Testcase caching is the source of recent bugs. I've decided to remove it and will re-think it later. Right now it doesn't work well with the zip testcase.


tool/clusterfuzz/binary_providers.py
```diff
@@ -235,7 +235,7 @@ def download_build_if_needed(dest, url):
   # args.gn is guaranteed to be in the wanted folder. In Chrome, it's under a
   # sub-directory. In Android, it's in the top dir.
   args_gn_path = find_file('args.gn', tmp_dir_path)
-  shutil.copy(os.path.dirname(args_gn_path), dest)
+  shutil.copytree(os.path.dirname(args_gn_path), dest)
```
inferno-chromium: Can you move os.path.dirname(args_gn_path) to a local and add comment on why we are using it. I don't understand why we do os.path.dirname, why don't we know the root of where testcase was unarchived.

tanin47: Oops, sorry, this part is not related to the testcase. This is a downloaded build. We've found that, for android, `args.gn` is in the root dir. But, for linux, `args.gn` is in a sub-folder.

PR #479, opened 2017-08-21
Android fixes

inferno-chromium:


tool/clusterfuzz/reproducers.py
```diff
@@ -660,6 +660,7 @@ def pre_build_steps(self):
   def reproduce_crash(self):
     """Reproduce crash on Android."""
     android.reset(self.testcase.android_package_name)
+    android.reboot()
```
tanin47: This will reboot on every iteration. Do we want to do that?

tool/tests/clusterfuzz/testcase_test.py
```diff
@@ -176,7 +176,7 @@ def test_get(self):
         {'content': 'debug.assert=0'},
         {'content': 'dalvik.vm.checkjni=true'},
         {'content': 'debug.checkjni=1'},
-        {'content': '[Environment] Command line file = /data/local/chrome-command-line with contents:'},
+        {'content': '[Environment] Command line file = /data/local/tmp/chrome-command-line with contents:'},
```
tanin47: Why do we need to add `tmp` here in the tests?

It doesn't seem matter. We just use whatever it is specified in the testcase. Do you mean that it should have been `/data/local/chrome-command-line`?

tool/clusterfuzz/android.py
```diff
@@ -234,6 +236,67 @@ def set_content_setting(table, key, value):
...
+    return expected in result.splitlines()
+
+  adb('wait-for-device')
+
+  start_time = time.time()
+  is_boot_completed = False
+  is_drive_ready = False
+  is_package_manager_ready = False
+  while time.time() - start_time < BOOT_TIMEOUT:
+    if not is_drive_ready:
```
tanin47: We can use a one-liner: `is_drive_ready = is_drive_ready or  drive_ready()`

tool/clusterfuzz/reproducers.py
inferno-chromium: Yes device settings reset wont work without reboot.

tool/tests/clusterfuzz/testcase_test.py
inferno-chromium: This never worked. We had COMMAND_LINE_PATH and COMMAND_LINE_PATH2 in job definition and COMMAND_LINE_PATH2 was the main one and COMMAND_LINE_PATH was the deprecated one. I fixed stacktraces for all testcases and removed COMMAND_LINE_PATH and renamed value in COMMAND_LINE_PATH2->COMMAND_LINE_PATH

tool/clusterfuzz/reproducers.py
tanin47: Cool, thanks.

tool/tests/clusterfuzz/testcase_test.py
tanin47: I see. Thanks.

PR #485, opened 2017-08-26
Gestures on Android

tanin47:
- Change `get_device_id` to setting the env variable instead. Otherwise, it wouldn't work with other adb commands.
- On my machine `adb shell getprop sys.boot_completed` returns `0\r\n`, so I fixed that part to be more robust.
- Implement gestures with monkey. I haven't reproduced any gesture testcase successfully yet. Not sure why.


tool/clusterfuzz/reproducers.py
```diff
@@ -217,15 +220,32 @@ def symbolize(output, source_dir_path):
...
+    return
+
+  gestures = gestures[0].split(',')
+
+  if gestures[0] != 'monkey':
+    return
+
+  android.adb_shell(
+      'monkey -p %s -s %s --throttle %s --ignore-security-exceptions %s' %
+      (package_name, gestures[-1], MONKEY_THROTTLE_DELAY, NUM_MONKEY_EVENTS))
```
inferno-chromium: move gestures[-1] to line 241
seed = gestures[1]

Issue #493, opened 2017-08-30
The CI can't run `sudo`


tanin47:
```
2017-08-30 20:09:18,060 [INFO]: [36mRunning: [1msudo[22m build/install-build-deps.sh --syms --no-prompt[39m
2017-08-30 20:09:18,063 [DEBUG]: ---------------------------------------
2017-08-30 20:09:18,071 [DEBUG]: sudo: no tty present and no askpass program specified
2017-08-30 20:09:18,071 [DEBUG]: Killing pid=30965 with 15
2017-08-30 20:09:18,072 [DEBUG]: ---------------------------------------
2017-08-30 20:09:18,072 [DEBUG]: | Return code is non-zero (1).
2017-08-30 20:09:18,072 [DEBUG]: | Exit.
2017-08-30 20:09:18,195 [INFO]: [33mCommandFailedError: `sudo build/install-build-deps.sh --syms --no-prompt` failed with the return code 1.[39m
```

tanin47:
It turns out that we need to specifically allow the user `clusterfuzz` to run `sudo` (not just its group).

PR #494, opened 2017-08-30
Allow user to passwordless sudo on CI

tanin47:
This fixes : #493


inferno-chromium:
can we remove the sudo from command start, or will it work fine for local runs ?

tanin47:
It works fine with the local run. The only change is that it'll ask for password at the beginning (as opposed to somewhere in the middel).

PR #500, opened 2017-09-05
Move job types to d8 standalone

inferno-chromium:


inferno-chromium:
see crbug.com/761654

for cfi, i think we need to create CFId8 builder ?

tool/clusterfuzz/resources/supported_job_types.yml
```diff
@@ -94,15 +95,13 @@ chromium:
         linux_lsan_chrome_mp: { preset: linux_chromium, sanitizer: LSAN }
         linux_msan_chrome: { preset: linux_chromium, builder: MsanChromium, sanitizer: MSAN }
         linux_msan_content_shell_drt: { preset: linux_content_shell, builder: MsanChromium, sanitizer: MSAN }
-        linux_msan_d8: { preset: linux_d8_chromium, builder: MsanChromium, sanitizer: MSAN }
```
tanin47: I'm a bit concerned about removing this. Because clusterfuzz uses a build with the whole chromium.

If you can only reproduce with standalone d8, it's not the exact match in environment.

Are we removing this chromium job?

tanin47:
> for cfi, i think we need to create CFId8 builder ?

It seems so. We only have chromium job type here: https://github.com/google/clusterfuzz-tools/pull/500/files#diff-3a6c61aba800675bbe1fcafcb20c67cfR92

tool/clusterfuzz/resources/supported_job_types.yml
inferno-chromium: We have already migrated these jobs to use standalone builds, check RELEASE_BUILD_URL_PATTERN in those job types.

PR #501, opened 2017-09-05
Add CFI V8 standalone job

tanin47:
- Use [the mixin pattern](https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-are-they-useful) instead. This eliminates a bunch of tests and redundant code.


tanin47:
Done. Thanks for pointing it out.

Issue #509, opened 2017-09-12
One crash is flaky but with `-i 10`... it reproduces.


tanin47:
Here's the crash: 4511494711803904

`-i 3` doesn't usually reproduce the crash. `-i 10` almost guarantees it.

PR #510, opened 2017-09-12
Increase iteration to 20

tanin47:
We've found one crash where CI doesn't reproduce.... but one of our users reproduces it locally after running it multiple times.

This fixes: #509 and increase our local reproducibility rate by 1%!


inferno-chromium:
LGTM.

PR #523, opened 2017-11-01
Fix for adb pull failures on non-library paths.

inferno-chromium:


tool/clusterfuzz/android.py
```diff
@@ -198,6 +198,9 @@ def find_lib_path(binary_path, search_paths, lib_tmp_dir_path):
   """Find the filename in search paths (or pull from the device) and return
     full path."""
   filename = os.path.basename(binary_path)
+  if not os.path.splitext(filename)[1] == ANDROID_LIBRARY_EXTENSION:
+    # Skip non-library paths.
+    return '<unknown>'
```
mbarbella: How is this value used? This seems a bit odd to me at a glance, but I don't remember much of the context for this.

inferno-chromium: The value is just ignored by symbolizer.

tool/clusterfuzz/android.py
```diff
@@ -198,6 +198,9 @@ def find_lib_path(binary_path, search_paths, lib_tmp_dir_path):
   """Find the filename in search paths (or pull from the device) and return
     full path."""
   filename = os.path.basename(binary_path)
+  if not os.path.splitext(filename)[1] == ANDROID_LIBRARY_EXTENSION:
```
tanin47: I wonder if we can guard the case where `binary_path` doesn't have an extension. Currently, it will throw `IndexError`.

inferno-chromium: No it wont.
a = ''
os.path.splitext(a)
('a', '')

tanin47: Well, TIL. Thank you.

PR #527, opened 2017-11-08
Hide access token

tanin47:
I can't merge it yet because my Pants is broken with the cryptography thing. I'll need to resolve that firs.t


tanin47:
Please don't merge yet. Let me fix the problem with cryptography first. Thanks.

PR #530, opened 2018-01-11
Upgrade pants to avoid crytography error

tanin47:
This avoids: 

```
13:23:22 00:00         [chroot]**** Failed to install cryptography-2.1.3 (caused by: NonZeroExit("received exit code 1 during execution of `[u'/usr/bin/python2.7', '-', 'bdist_wheel', '--dist-dir=/tmp/tmp68gqf2']`",)
):
stdout:

stderr:
error in cryptography setup command: Invalid environment marker: python_version < '3'
```

When running test.

However, the new version of Pants seems to use a broken version of pytest. See: https://github.com/pantsbuild/pants/issues/5310 -- when running test, we need to add `-- -p no:logging`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,5 +1,6 @@
 dependencies:
   override:
+    - sudo apt-get update
     - sudo apt-get install libpython2.7-dev
     - ./pants -V
   cache_directories:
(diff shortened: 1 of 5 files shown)


tanin47:
There is some weird thing going on with coverage.py. I'll need to fix that problem before merging.

inferno-chromium:
Thanks a lot for taking the time to fix this.

PR #534, opened 2018-11-06
Update README.md

jonathanmetzman:
Note that this tool does not work with libFuzzer on Windows.
Link to docs explaining how to handle this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@ ClusterFuzz tools
 The tools supports various tasks (e.g. reproduce a crash locally)
 needed by ClusterFuzz's users.
 
+It works on Linux and Mac. 
+Follow the instructions on [Reproducing on Windows](https://chromium.googlesource.com/chromium/src/+/master/testing/libfuzzer/reproducing_on_windows.md) to reproduce a crash found by libFuzzer on Windows.
+
 Currently, it supports reproducing a crash locally. In the future, it will
 support uploading a fuzzer, tailing fuzzer log, and uploading a testcase.
 


jonathanmetzman:
PTAL.

README.md
```diff
@@ -7,6 +7,9 @@ ClusterFuzz tools
 The tools supports various tasks (e.g. reproduce a crash locally)
 needed by ClusterFuzz's users.
 
+It works on Linux and Mac. 
+Follow the instructions on [Reproducing on Windows](https://chromium.googlesource.com/chromium/src/+/master/testing/libfuzzer/reproducing_on_windows.md) to reproduce a crash found by libFuzzer on Windows.
```
Dor1s: Would it be better to suggest users to follow the link from the issue description instead?

inferno-chromium: Lets do both. i think we should add this in job type definition as well so that correct link is shown directly.

jonathanmetzman: Which issue description? The issue filed for a CF crash? I think that would be good, but I want to make sure people who come across this (or have used it in the past) don't get confused.

jonathanmetzman: OK. Considering this resolved. Is this fine to land?

Dor1s: Yes, a Monorail issue. Can we add one more sentence / paragraph after this like "In other cases, please refer to the issue description for the most accurate information on crash reproducing"?

Dor1s: Looks like race condition between out comments, re-opening for my last one.

jonathanmetzman:
I also changed the verbs to agree with "tools" which is plural

README.md
jonathanmetzman: Added the line suggested by Max and mentioned Android.

PR #543, opened 2018-12-20
Support new jobs

jonathanmetzman:
Support the following new jobs:

x86_libfuzzer_chrome_asan
x86_libfuzzer_chrome_asan_debug

### DIFF
--- a/tool/clusterfuzz/resources/supported_job_types.yml
+++ b/tool/clusterfuzz/resources/supported_job_types.yml
@@ -75,8 +75,11 @@ standalone:
         linux_cfi_d8: { preset: d8_standalone, builder: CfiV8, sanitizer: CFI }
 chromium:
         afl_chrome_asan: { preset: afl, sanitizer: ASAN }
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I'm actually unsure if we should support chromeos_libfuzzer_chrome_asan, since developers who use it need to have "chromeos" in their `target_os` list in their Chrome repo's `.gclient` file.

jonathanmetzman:
Going to land since it looks like the tool already "supports: `linux_asan_chrome_chromeos` which has the same requirement. Though I think it's somewhat wrong that we support it, since it won't "just work" for most devs (they will need to add "chromeos" to their `target_os` list in Chrome's `.gclient` file).

jonathanmetzman:
I think the format makes sense in that we have chromeos_libfuzzer_chrome_asan, windows_libfuzzer_chrome_asan, and Mac, so I assumed it makes sense to also use this as prefix.

jonathanmetzman:
Anyway, I'll keep x86_libfuzzer_chrome_asan experimental until you decide if this needs to be migrated.

PR #545, opened 2018-12-23
Modernize Python 2 code to get ready for Python 3

cclauss:
Use __print()__ function in both Python 2 and Python 3
* Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

Old style exceptions --> new style for Python 3
* Old style exceptions are syntax errors in Python 3 but new style exceptions work as expected in both Python 2 and Python 3.

### DIFF
--- a/tool/clusterfuzz/output_transformer.py
+++ b/tool/clusterfuzz/output_transformer.py
@@ -1,5 +1,10 @@
 """Transform the output before printing on screen."""
 
+try:
+  xrange
+except NameError:
+  xrange = range
+
 
 class Base(object):
   """Transform output and send to the output function."""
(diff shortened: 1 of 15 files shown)


tool/clusterfuzz/output_transformer.py
```diff
@@ -1,5 +1,10 @@
 """Transform the output before printing on screen."""
 
+try:
+  xrange
+except NameError:
+  xrange = range
```
Dor1s: Let's just use `range` in this file as well, there should not be too many elements.

Dor1s: Makes sense. Closing this PR then. Sorry, @cclauss.

PR #547, opened 2019-01-10
Add v8 simulator jobs

jonathanmetzman:


jonathanmetzman:
Abhishek could you please run `./pants run butler -- release` on trunk? I don't have permission to write to the x20 directory.
Thanks!

PR #549, opened 2019-04-09
Handle whitespace around "=" in environment variable parsing.

mbarbella-chromium:
Also updates pants version to avoid Python 2 compatibility issues with newer versions of pytest.


tool/tests/clusterfuzz/testcase_test.py
```diff
@@ -88,6 +88,32 @@ def test_parsing_json(self):
...
+        {'content': '[Environment] LSAN_OPTIONS=y=1'},
+        {'content': ('Running command: /path/to/binary --random-seed=23 '
+                     '--turbo /path/to/testcase')},
+        {'content': '[Environment] TEST_TWO=third=3:fourth=4'}]
+    result = build_base_testcase(
+        stacktrace_lines=stacktrace_lines, revision=5, build_url='build_url',
+        gestures=True)
+    self.assertEqual(result.id, '12345')
+    self.assertEqual(result.revision, 5)
+    self.assertEqual(result.environment, {'TEST_ARGS': 'first=1:second = 2',
```
inferno-chromium: Is there a test on if second is parsed correctly and is assigned value 2 (without spaces ?)

mbarbella-chromium: The difference was the whitespace around "TEST_ARGS = ..." vs "TEST_ARGS=...". There was actually an explicit test above that it retains the second set of whitespace in "second = 2" so I didn't want to change this behavior. I think that's fine, since that part is meant to simulate something like colon delimited asan options which shouldn't actually have whitespace.

inferno-chromium: Thanks!

PR #550, opened 2019-04-09
Ignore quotes around environment variable strings.

mbarbella-chromium:


tool/clusterfuzz/testcase.py
```diff
@@ -59,6 +59,10 @@ def get_environment_and_args(stacktrace_lines):
         continue
       name, value = [token.strip() for token in tokens]
 
+      # Strip quotes around environment variable values if necessary.
```
inferno-chromium: nit: double quotes.

